PR #12, opened 2015-04-02
Add compression support in output plugins

takebayashi:
To support Kafka compression in out_kafka and out_kafka_bufferd, add a new configuration parameter `compression_codec` and pass the value to Poseidon's producers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,7 @@ Or install it yourself as:
       max_send_retries    (integer)    :default => 3
       required_acks       (integer)    :default => 0
       ack_timeout_ms      (integer)    :default => 1500
+      compression_codec   (none|gzip|snappy) :default => none
     </match>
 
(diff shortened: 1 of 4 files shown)


htgc:
Thank you for your contribution!
Can you update README.md and fluent-plugin-kafka.gemspec also?

takebayashi:
Added the description about `compression_codec` to README.md and the dependency for snappy gem to fluent-plugin-kafka.gemspec.

htgc:
I removed 'snappy' from fluent-plugin-kafka.gemspec because some user might failed updating this plugin due to lack of snappy library.
i.e. User should install snappy library before update this module, and it might be troublesome.

Therefore, I added instructions to install snappy into README.me.
Please let me know if you have any problem and sorry for inconvenience.

htgc:
Fixed #12

PR #13, opened 2015-04-04
Add formatter plugin support by calling formatter as a fallback for unknown output_data_type

takebayashi:
The out_kafka and out_kafka_buffered plugins have their own implementations for multiple output formats (w/ `output_data_type` option), but not have supports for Fluentd formatter plugins.

To support formatter plugins, make output plugins create formatter plugin instance and call its `format` method when unknown value is passed at `output_data_type`. 

For example, when the following configuration is passed, the output plugin uses `CsvFormatter` plugin to format the output data.
```
output_data_type csv
fields message
force_quotes false
```

This change has a potential to break backward compatibilities if the user already passes the unknown value to `output_data_type`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,7 +41,7 @@ Or install it yourself as:
       zookeeper           <zookeeper_host>:<zookeeper_port> # Set brokers via Zookeeper
 
       default_topic       <output topic>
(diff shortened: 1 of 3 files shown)


summerwind:
+1
I like this feature.

PR #23, opened 2015-08-28
Create a hash for input text message

Jecht:
Should create a hash when input record is in plain text, otherwise the record
will be discarded by output plugins.  This was found in our use case: extract
plain text messages from kafka and output to elasticsearch.

See also: [fluentd/fluentd/issues/664](https://github.com/fluent/fluentd/issues/664#issuecomment-135394548).


ymattw:
Hi @htgc, this fix is very simple and straightforward, looks perfect to me.  Hope you can take some time to review. Thanks.

htgc:
sorry for late reply and thank you for your contribution!
I've merged and released it.

Issue #24, opened 2015-09-06
Feature request: use poseidon_cluster to support consumer group and zookeeper


ymattw:
Hi @htgc, thanks for your great work on this plugin, I am working on a log aggregation solution with kafka and fluentd, I would like to know, do you have any plan to support consumer group and zookeeper in the kafka input plugin

I have tried [poseidon_cluster](https://github.com/bsm/poseidon_cluster), with minimal change I made it work in my environment.  It requires more work to be solid but it's not a big change.  If you want I can work on a pull request, just want to touch base first in case you have the plan.  Thanks!

ymattw:
Greetings @htgc, I have a working version ready at [in_kafka_group.rb](https://github.com/ymattw/fluent-plugin-kafka/blob/consumer-group/lib/fluent/plugin/in_kafka_group.rb).  The idea is to add a new plugin `in_kafka_group` which leverages [poseidon_cluster](https://github.com/bsm/poseidon_cluster), I choose to implement new one because its configuration is not backward compatible.

If this works for you, I can proceed complete other changes like gemspec and documentation.  Please take a look and let me know.  Thanks.

htgc:
Hello @ymattw , and thank you for your contribution!
I have already implemented output plugin to get broker list from zookeeper if user set `zookeeper` parameter in configuraiton.
I'd like to merge in_kafka_group feature (i.e. switch to use poseidon_cluster instead of poseidon) into in_kafka if user set `zookeeper` in configuration as same as output plugin.
Please let me know your opinion.
Thanks,

ymattw:
That's awesome, I will be pleased as long as **consumer group** is supported as well as zookeeper.  Thanks!

ymattw:
Hey @htgc, I have maintained my own branch for this feature for a while, I am looking forward this to be implemented in your next release, any roadmap so far?  Thanks!

htgc:
Hello @ymattw,
Sorry for inconvenience. Actually, I didn't have plan.
I would be happy if you can implement as above but please let me know if you can't. I will plan to do that.
Thanks,

ymattw:
Hi @htgc,I wouldn't  merge `in_kafka_group` to `in_kafka` because that will be a major change and user impacting, but it's totally up to you :), and please feel free to reuse my prototype mentioned earlier.

PS: The `in_kafka_group` implementation have been working well in prod so far, thanks again for your work in in_kafka.

htgc:
[...]

PR #32, opened 2015-11-02
Consumer group

ymattw:
This is a follow up to fix issue #24.


htgc:
Could you add `poseidon_cluster` into fluent-plugin-kafka.gemspec? Thanks!

ymattw:
My bad, just pushed a new commit.

PR #53, opened 2016-05-30
Adopt v0.14

cosmo0920:
Hi, I've tried to adopt Fluentd v0.14.

I've found that there are several problems and I fixed them.

* missing development dependency of test-unit gem
* missing `fluent/input` require
* sometimes missing calling `super` in `#start` and `#shutdown`


cosmo0920:
Thanks!

PR #55, opened 2016-06-09
Removed the `zookeeper_path` option from the input config.

sww:
The input plugin currently doesn't support `zookeeper_path`.  This was a source of confusion when trying out the plugin.


repeatedly:
Thx!

PR #57, opened 2016-07-05
[WIP] Improve performance and reliability

repeatedly:
This patch introduce two changes
- Add TopicProducer to reduce ruby-kafak CPU usage
- Separate chunks into each topic for reliable data flush in `write` method

Tested with 20000msgs/sec with single_value format, CPU usage is reduced from around 65% to 45%.

### DIFF
--- a/lib/fluent/plugin/out_kafka_buffered.rb
+++ b/lib/fluent/plugin/out_kafka_buffered.rb
@@ -8,6 +8,7 @@ class Fluent::KafkaOutputBuffered < Fluent::BufferedOutput
     super
 
     require 'kafka'
+    require 'fluent/plugin/kafka_ext'
 
     @kafka = nil
     @producers = {}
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_kafka_buffered.rb
```diff
@@ -143,8 +144,21 @@ def shutdown
...
+    @emit_count += 1
+    events = {}
+    def_topic = @default_topic || tag
+    es.each { |time, record|
+      topic = record['topic'] || def_topic
+      events[topic] ||= ''.force_encoding('ASCII-8BIT')
+      events[topic] << [tag, time, record].to_msgpack
+    }
+    events.each { |topic, data|
+      # TODO: How to handle BufferQueueLimitError after stored several events?
```
repeatedly: One way is using error stream to route such events.

PR #62, opened 2016-07-20
Use ruby-kafka instead of poseidon in out_kafka

repeatedly:
Next step for removing poseidon dependency.


repeatedly:
@htgc If you okay, I will merge this.

htgc:
LGTM! thanks

PR #68, opened 2016-08-25
Update readme

kentarosasaki:
This PR is going to update READ.md.

* Poseidon is already not being used and ruby-kafka is used instead of Poseidon.
* In ruby-kafka, the simplest option is to not specify a message key, partition key, or partition number, in which case the message will be assigned a partition at random.


repeatedly:
Thanks!
I forgot to update several sentence...

PR #98, opened 2016-11-10
Allow specifying the partition to use

jcalvert:
The same way that is done for specifying the `partition_key` to use by embedding it in the message and then removing it, this pull request allows a client to do so with `partition` as well. Since the `ruby-kafka` client supports specifying the partition as an alternative to `partition_key`, this is a simple change and remains consistent with the `ruby-kafka` library.


jcalvert:
Bump for feedback!

repeatedly:
Sorry. I forgot to merge this... :p
I will release new version soon.

repeatedly:
done

jcalvert:
🙇

PR #102, opened 2017-01-06
Support for optional replacing of record's timestamp

bohyn:
Support for optional replacing of record's timestamp by value of 'time' field.

### DIFF
--- a/lib/fluent/plugin/in_kafka.rb
+++ b/lib/fluent/plugin/in_kafka.rb
@@ -54,6 +54,8 @@ class Fluent::KafkaInput < Fluent::Input
   def initialize
     super
     require 'kafka'
+
+    @time_parser = nil
   end
 
   def configure(conf)
@@ -101,6 +103,10 @@ class Fluent::KafkaInput < Fluent::Input
     require 'zookeeper' if @offset_zookeeper
 
     @parser_proc = setup_parser
+
+    if @use_record_time and @time_format
+      @time_parser = Fluent::TextParser::TimeParser.new(@time_format)
+    end
   end
 
   def setup_parser
@@ -239,8 +245,7 @@ class Fluent::KafkaInput < Fluent::Input
           record = @parser.call(msg, @topic_entry)
           if @use_record_time
             if @time_format
-              time_parser = Fluent::TextParser::TimeParser.new(@time_format)
-              record_time = time_parser.parse(record['time'])
+              record_time = @time_parser.parse(record['time'])
             else
               record_time = record['time']
             end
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_kafka.rb
```diff
@@ -31,6 +31,10 @@ class Fluent::KafkaInput < Fluent::Input
 
   config_param :offset_zookeeper, :string, :default => nil
   config_param :offset_zk_root_node, :string, :default => '/fluent-plugin-kafka'
+  config_param :input_replace_time, :bool, :default => false,
```
repeatedly: `use_record_time` is more clear

lib/fluent/plugin/in_kafka.rb
```diff
@@ -231,7 +235,16 @@ def consume
 
       messages.each { |msg|
         begin
-          es.add(Fluent::Engine.now, @parser.call(msg, @topic_entry))
+          record = @parser.call(msg, @topic_entry)
+          record_time = Fluent::Engine.now
+          if @input_replace_time
+            if @time_format
+              record_time = Time.strptime(record['time'], @time_format)
```
repeatedly: Don't use `Time.strptime`.
Use Fluentd's `TimeParser` instead.

lib/fluent/plugin/in_kafka.rb
```diff
@@ -231,7 +235,16 @@ def consume
 
       messages.each { |msg|
         begin
-          es.add(Fluent::Engine.now, @parser.call(msg, @topic_entry))
+          record = @parser.call(msg, @topic_entry)
+          record_time = Fluent::Engine.now
```
repeatedly: Don't create `Fluent::Engine.now` when option is true.
It allocates unnecessary object.

lib/fluent/plugin/in_kafka.rb
```diff
@@ -231,7 +236,18 @@ def consume
 
       messages.each { |msg|
         begin
-          es.add(Fluent::Engine.now, @parser.call(msg, @topic_entry))
+          record = @parser.call(msg, @topic_entry)
+          if @use_record_time
+            if @time_format
+              time_parser = Fluent::TextParser::TimeParser.new(@time_format)
```
repeatedly: Creating TimeParser in `each` is very high cost.
It should be created in `configure` when `use_record_time` is `true`

bohyn: Fixed

repeatedly:
Thanks!

Issue #108, opened 2017-02-06
max_bytes option in in_kafka_group is not working


withsmilo:
I would like to change max bytes of fetched data to larger, but max_bytes option is not working. Is this bug?

withsmilo:
[ruby-kafka's consumer](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer.rb#L83) has a max_bytes option set by 1M default. I fixed this bug. I will create a pull request tonight. :smile: 
```
[in_kafka_group.rb]
.....
  config_param :max_bytes, :integer, :default => nil,
               :desc => "Maximum bytes to fetch"
.....
  def setup_consumer
    consumer = @kafka.consumer(@consumer_opts)
    @topics.each { |topic|
      consumer.subscribe(topic, start_from_beginning: @start_from_beginning, max_bytes_per_partition: @max_bytes)
    }
    consumer
  end
.....
```

repeatedly:
Yeah, it's a README mistake.
I just removed from kafka_group.

> I will create a pull request tonight. 

Thanks. i wait it :)

PR #110, opened 2017-02-06
Add max_bytes option to in_kafka_group

withsmilo:
* Related issue is #108.
* ruby-kafka's subscribe function supports max_bytes_per_partition parameter.
  * refer to [link](https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/consumer.rb#L83)
  * default value is 1048576(1M).


repeatedly:
Thanks!

repeatedly:
Released v0.5.2

Issue #120, opened 2017-04-10
[out-kafka] Kafka Buffer Overflow when event stream has more than 1000 messages


yoelcabo:
Since the plugin uses the doesn't call `producer.deliver_messages` until it has `producer.produce` every message, the buffer overflows. Besides, the size of the buffer is not configurable.

```
emit transaction failed: error_class=Kafka::BufferOverflow error="Cannot produce to <mytopic>, max buffer size (1000 messages) reached" tag="<mytag>"
```

I am going to send right now two pull requests to solve that, in my opinion both should be merged.

repeatedly:
Merged patches.

PR #122, opened 2017-04-10
make parameters of ruby-kafka configurable

yoelcabo:
Makes easy to mitigate https://github.com/fluent/fluent-plugin-kafka/issues/120.

Makes the parameters max_buffer_size and max_buffer_bytesize of ruby-kafka configurable.

### DIFF
--- a/lib/fluent/plugin/out_kafka_buffered.rb
+++ b/lib/fluent/plugin/out_kafka_buffered.rb
@@ -62,6 +62,11 @@ DESC
 The codec the producer uses to compress messages.
 Supported codecs: (gzip|snappy)
 DESC
+  config_param :max_buffer_size, :integer, :default => nil,
+	                 :desc => "Number of messages to be buffered by the kafka producer."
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_kafka_buffered.rb
```diff
@@ -62,6 +62,11 @@ class Fluent::KafkaOutputBuffered < Fluent::BufferedOutput
 The codec the producer uses to compress messages.
 Supported codecs: (gzip|snappy)
 DESC
+  config_param :max_buffer_size, :integer, :default => nil,
```
repeatedly: These parameters are not needed because out_kafka_buffered and out_kafka2 don't use original `produce` method.

yoelcabo: In this case, let's do the changes to out_kafka only.

repeatedly:
Thanks!

PR #126, opened 2017-04-18
Add discard_kafka_delivery_failed option to handle Kafka::DeliveryFailed

wyukawa:
Hi, I use Kafka 0.9 and fluent-plugin-kafka 0.5.4.

If user uses a period ('.') and underscore ('_') as kafka topic name, it's troublesome.

For example, user creates a.a_a topic
```
$ bin/kafka-topics.sh --create --zookeeper localhost:2181 --replication-factor 1 --partitions 1 --topic a.a_a
WARNING: Due to limitations in metric names, topics with a period ('.') or underscore ('_') could collide. To avoid issues it is best to use either, but not both.
Created topic "a.a_a".
```

but, user sends message to a.a.a topic.
```
  <source>
    @type forward
  </source>
  <match a.a.a>
    type kafka_buffered
    brokers localhost:9092
    flush_interval 1
  </match>
```

```
$ echo '{"hoge":"fuga"}' | fluent-cat a.a.a
```

Error occurs as follows
```
2017-04-18 10:52:43 +0900 [warn]: temporarily failed to flush the buffer. next_retry=2017-04-18 10:53:51 +0900 error_class="Kafka::DeliveryFailed" error="Failed to assign partitions to 1 messages" plugin_id="object:3fe35d43b5cc"
  2017-04-18 10:52:43 +0900 [warn]: suppressed same stacktrace
2017-04-18 10:53:53 +0900 [warn]: Send exception occurred: Failed to assign partitions to 1 messages
2017-04-18 10:53:53 +0900 [warn]: Exception Backtrace : /.../vendor/bundle/gems/ruby-kafka-0.3.17/lib/kafka/producer.rb:325:in `deliver_messages_with_retries'
/.../vendor/bundle/gems/ruby-kafka-0.3.17/lib/kafka/producer.rb:243:in `block in deliver_messages'
/.../vendor/bundle/gems/ruby-kafka-0.3.17/lib/kafka/instrumenter.rb:21:in `instrument'
/.../vendor/bundle/gems/ruby-kafka-0.3.17/lib/kafka/producer.rb:236:in `deliver_messages'
/.../vendor/bundle/gems/fluent-plugin-kafka-0.5.4/lib/fluent/plugin/out_kafka_buffered.rb:271:in `write'
/.../vendor/bundle/gems/fluentd-0.12.33/lib/fluent/buffer.rb:354:in `write_chunk'
/.../vendor/bundle/gems/fluentd-0.12.33/lib/fluent/buffer.rb:333:in `pop'
/.../vendor/bundle/gems/fluentd-0.12.33/lib/fluent/output.rb:342:in `try_flush'
/.../vendor/bundle/gems/fluentd-0.12.33/lib/fluent/output.rb:149:in `run'
```

User may create a troublesome kafka topic because user can create fluentd tag(kafka topic) freely in my environment.

IMHO, it's not necessary to retry if Kafka::DeliveryFailed error occurs because retry doesn't succeed and increases fluentd buffer size.

So, I want to skip such kind a record.

I want to add a option(skip_kafka_delivery_failed) because fluent-plugin-kafka become more robust.
```
  <source>
    @type forward
  </source>
  <match a.a.a>
    type kafka_buffered
    brokers localhost:9092
    flush_interval 1
    skip_kafka_delivery_failed true
  </match>
 ```
 
 log
 ```
2017-04-18 13:02:49 +0900 [warn]: DeliveryFailed occurred. Skip broken event: error="Failed to assign partitions to 1 messages" error_class="Kafka::DeliveryFailed" tag="a.a.a"
 ```

Default setting of skip_kafka_delivery_failed is false, so I think that there is no regression for existing users.

### DIFF
--- a/lib/fluent/plugin/out_kafka_buffered.rb
+++ b/lib/fluent/plugin/out_kafka_buffered.rb
@@ -63,6 +63,7 @@ The codec the producer uses to compress messages.
 Supported codecs: (gzip|snappy)
 DESC
   config_param :max_send_limit_bytes, :size, :default => nil
+  config_param :skip_kafka_delivery_failed, :bool, :default => false
 
   config_param :time_format, :string, :default => nil
 
@@ -209,6 +210,19 @@ DESC
     end
   end
 
+  def deliver_messages(producer, tag)
+    if @skip_kafka_delivery_failed
+      begin
+        producer.deliver_messages
+      rescue Kafka::DeliveryFailed => e
+        log.warn "DeliveryFailed occurred. Skip broken event:", :error => e.to_s, :error_class => e.class.to_s, :tag => tag
+        producer.clear_buffer
+      end
+    else
+      producer.deliver_messages
+    end
+  end
+
   def write(chunk)
     tag = chunk.key
     def_topic = @default_topic || tag
@@ -254,7 +268,7 @@ DESC
 
         if (messages > 0) and (messages_bytes + record_buf_bytes > @kafka_agg_max_bytes)
           log.debug { "#{messages} messages send because reaches kafka_agg_max_bytes" }
-          producer.deliver_messages
+          deliver_messages(producer, tag)
           messages = 0
           messages_bytes = 0
         end
@@ -268,7 +282,7 @@ DESC
       }
       if messages > 0
         log.debug { "#{messages} messages send." }
-        producer.deliver_messages
+        deliver_messages(producer, tag)
       end
       log.debug { "(records|bytes) (#{records_by_topic}|#{bytes_by_topic})" }
     end
--- a/lib/fluent/plugin/out_kafka_buffered.rb
+++ b/lib/fluent/plugin/out_kafka_buffered.rb
@@ -63,7 +63,7 @@ The codec the producer uses to compress messages.
 Supported codecs: (gzip|snappy)
 DESC
   config_param :max_send_limit_bytes, :size, :default => nil
-  config_param :skip_kafka_delivery_failed, :bool, :default => false
+  config_param :discard_kafka_delivery_failed, :bool, :default => false
 
   config_param :time_format, :string, :default => nil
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by repeatedly** [comment]
I didn't check all exception classes.
`DeliveryFailed` is for unexpected situation?
Does network related failure return other class?

**Comment by wyukawa** [comment]
hmm, ruby-kafka raises DeliveryFailed when it's buffer is not empty.
https://github.com/zendesk/ruby-kafka/blob/v0.3.17/lib/kafka/producer.rb#L322-L333

It seems to be OK if fluent-plugin-kafka doesn't retry when DeliveryFailed error occured because ruby-kafka has also retry logic.

Network related failure returns Kafka::ConnectionError.

File: lib/fluent/plugin/out_kafka_buffered.rb [hunk 1]
```diff
@@ -63,6 +63,7 @@ class Fluent::KafkaOutputBuffered < Fluent::BufferedOutput
 Supported codecs: (gzip|snappy)
 DESC
   config_param :max_send_limit_bytes, :size, :default => nil
+  config_param :skip_kafka_delivery_failed, :bool, :default => false
```
repeatedly: Logging warning message in `configure` is better when this option is true.
"This option discards events which cause delivery failure, e.g. invalid topic or something. If this is unexpected, you need to check your configuration or data"

lib/fluent/plugin/out_kafka_buffered.rb
```diff
@@ -209,6 +210,19 @@ def setup_formatter(conf)
     end
   end
 
+  def deliver_messages(producer, tag)
+    if @skip_kafka_delivery_failed
+      begin
+        producer.deliver_messages
+      rescue Kafka::DeliveryFailed => e
+        log.warn "DeliveryFailed occurred. Skip broken event:", :error => e.to_s, :error_class => e.class.to_s, :tag => tag
```
repeatedly: `discard` is better than `skip`.

lib/fluent/plugin/out_kafka_buffered.rb
wyukawa: OK, I add

lib/fluent/plugin/out_kafka_buffered.rb
wyukawa: changed

wyukawa:
Thank you for comments!

I rebased to resolve conflict and applied your comments.

repeatedly:
Thanks! Just merged.

wyukawa:
Thanks!

PR #134, opened 2017-06-23
Add optional emit for active support notifications

jcalvert:
This pull request allows a user of the plugin to specify a regular expression to subscribe to `ActiveSupport::Notifications` and then re-emit those notifications for processing downstream, such as with with `<match>` directives in the Fluentd configuration. This is useful because the Ruby Kafka gem [already supports ActiveSupport notifications](https://github.com/zendesk/ruby-kafka#instrumentation) allowing users to gather statistics and information.

### DIFF
--- a/lib/fluent/plugin/out_kafka.rb
+++ b/lib/fluent/plugin/out_kafka.rb
@@ -60,6 +60,12 @@ DESC
   config_param :max_buffer_bytesize, :integer, :default => nil,
                :desc => "Maximum size in bytes to be buffered."
 
+  config_param :active_support_notification_regex, :string, :default => nil,
+               :desc => <<-DESC
+Add a regular expression to capture ActiveSupport notifications from the Kafka client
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_kafka.rb
```diff
@@ -60,6 +60,8 @@ class Fluent::KafkaOutput < Fluent::Output
   config_param :max_buffer_bytesize, :integer, :default => nil,
                :desc => "Maximum size in bytes to be buffered."
 
+  config_param :active_support_notification_regex, :string, :default => nil
```
repeatedly: Add `:desc` parameter

jcalvert:
@repeatedly Updated with description and using the router to more gracefully `emit` the message instead of just using `log`. Thank you!

jcalvert:
bump please

repeatedly:
Sorry. Just released v0.5.6.

jcalvert:
Thank you! 🙇

PR #137, opened 2017-07-10
Add kafka_agg_max_messages option

kanga333:
Hi.
I'd like to add an option for batch transfer, which sets an upper limit on the number of messages.
Currently, `kafka_agg_max_bytes` can set an upper limit on the number of bytes, but we also want to limit the number of messages.
If only `kafka_agg_max_bytes` is used, there is a possibility that a lot of small logs are packed in one batch.
The number of logs in one batch affects the number of duplicate logs at retransmission, so I would like to control this with the parameter `kafka_agg_max_messages`.
Also, `kafka_agg_max_bytes` is an important parameter affecting throughput, but there is no explanation in the `README.md`. I think that adding the explanation to the `README.md` is more helpful to the user.


repeatedly:
Thanks!

kanga333:
Thank you for the merge.

Issue #139, opened 2017-07-26
Feature Request - support Statsd and Datadog monitoring


wyukawa:
ruby-kafka supports  Statsd and Datadog monitoring.
https://github.com/zendesk/ruby-kafka#monitoring
So, if fluent-plugin-kafka supports such kind of monitorings, it would be nice.

Thanks

PR #140, opened 2017-07-26
Add statsd and datadog monitoring

wyukawa:
#139 

ruby-kafka supports Statsd and Datadog monitoring.
https://github.com/zendesk/ruby-kafka#monitoring

ruby-kafka 4.0 supports statsd.
https://github.com/zendesk/ruby-kafka/pull/373

This pull request enables to monitor kafka client by using datadog or statsd

sample fluentd conf
```
  <source>
    @type forward
  </source>
  <match aaa>
    type kafka_buffered
    statsd_monitoring true
    get_kafka_client_log true
    brokers localhost:9092
    flush_interval 1
  </match>
```

sample Gemfile
```
source 'https://rubygems.org'

gem "fluent-plugin-kafka",:git=>"https://github.com/wyukawa/fluent-plugin-kafka.git", :ref => "281f1b66cd303ea8b16333364dba21771bbad922"
gem "statsd-ruby"
gem "activesupport", "5.1.2"
gem "fluentd", "0.12.33"
```

I confirmed by using https://github.com/prometheus/statsd_exporter

I can't confirm datadog monitoring because I don't use datadog.

Thanks

### DIFF
--- a/lib/fluent/plugin/out_kafka_buffered.rb
+++ b/lib/fluent/plugin/out_kafka_buffered.rb
@@ -74,6 +74,9 @@ Add a regular expression to capture ActiveSupport notifications from the Kafka c
 requires activesupport gem - records will be generated under fluent_kafka_stats.**
 DESC
 
+  config_param :statsd_monitoring, :bool, :default => false
+  config_param :datadog_monitoring, :bool, :default => false
+
   include Fluent::KafkaPluginUtil::SSLSettings
   include Fluent::KafkaPluginUtil::SaslSettings
 
@@ -166,6 +169,16 @@ DESC
         message = event.payload.respond_to?(:stringify_keys) ? event.payload.stringify_keys : event.payload
         @router.emit("fluent_kafka_stats.#{event.name}", Time.now.to_i, message)
       end
+    end
+
+    if @statsd_monitoring
+      require 'kafka/statsd'
+      log.info 'statsd monitoring started'
+    end
+
+    if @datadog_monitoring
+      require 'kafka/datadog'
+      log.info 'datadog monitoring started'
     end
   end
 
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_kafka_buffered.rb
```diff
@@ -74,6 +74,9 @@ class Fluent::KafkaOutputBuffered < Fluent::BufferedOutput
 requires activesupport gem - records will be generated under fluent_kafka_stats.**
 DESC
 
+  config_param :statsd_monitoring, :bool, :default => false
```
repeatedly: How about using array type parameter instead of each single parameter like below?

```
config_param :monitoring_list, :array, :default => []
```

With this approach, we can write configuration simply.

```
@monitoring_list.each { |m|
  require "kafka/#{m}"
  log.info "..."
}
```

wyukawa:
Thank you for comment!

To use array is better.
So, I fixed https://github.com/fluent/fluent-plugin-kafka/pull/140/commits/0a90f59f48ad6449536d3554037e66b1c8a5751f

lib/fluent/plugin/out_kafka_buffered.rb
```diff
@@ -74,6 +74,8 @@ class Fluent::KafkaOutputBuffered < Fluent::BufferedOutput
 requires activesupport gem - records will be generated under fluent_kafka_stats.**
 DESC
 
+  config_param :monitoring_list, :array, :default => []
```
repeatedly: Could you add `:desc` parameter?

wyukawa: added b97561ff99e02060b929e3ad69d0ebcc72d92c82

wyukawa:
Thanks!

PR #155, opened 2017-10-26
Re-starting the consumer in case of network failer or kafka brokers a…

sharadgaur:
I am restarting the consumer in case Kafka consumer was not able to recover/re-connect/re-consume in case of network outage between consumer and broker or kafka brokers got restarted.

Also added the new config variable retry_wait_seconds (default value 30) for delay.

~Sharad

### DIFF
--- a/lib/fluent/plugin/in_kafka_group.rb
+++ b/lib/fluent/plugin/in_kafka_group.rb
@@ -191,7 +191,6 @@ class Fluent::KafkaGroupInput < Fluent::Input
             rescue => e
               log.warn "parser error in #{batch.topic}/#{batch.partition}", :error => e.to_s, :value => msg.value, :offset => msg.offset
               log.debug_backtrace
-              reconnect_consumer
             end
           }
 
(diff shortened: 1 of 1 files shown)


sharadgaur:
@repeatedly Please have a look and let me know if I need to change something. We need this change in production asap.

Thanks,
Sharad

lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -176,6 +191,7 @@ def run
             rescue => e
               log.warn "parser error in #{batch.topic}/#{batch.partition}", :error => e.to_s, :value => msg.value, :offset => msg.offset
               log.debug_backtrace
+              reconnect_consumer
```
repeatedly: Is this point correct for reconnecting?
This `rescue` is for catching parsing error so kafka brokers should not happen this point.

sharadgaur: oh good catch .. let me fix and re-test it

sharadgaur:
@repeatedly  Please review.

sharadgaur:
@repeatedly  Please review it as soon as possible. We need these changes in prod.

Thanks,
Sharad

PR #162, opened 2017-11-18
add config for multi-worker support

lee2014:
When I set multi_worker configuration with in_udp and out_kafka, it throws such exception: 
"error_class=Fluent::ConfigError error="Plugin 'kafka_buffered' does not support multi workers configuration". So I add a parameter for kafka's plugins to enable multi_worker_ready.


repeatedly:
Why not always returns `true`?
I think `config_param` is not needed?

lee2014:
@repeatedly  You are right. There is no need for this configuration. 
And in_kafka doesn't support multi_worker mode.

repeatedly:
Thx!

PR #168, opened 2017-12-18
Added support for PLAIN/SCRAM SASL

roche-d:
Simple patch to support PLAIN/SCRAM authentication


repeatedly:
Thanks for the patch!
But the copy & paste is hard to maintain.
How about this client setup routine into here?

https://github.com/fluent/fluent-plugin-kafka/blob/master/lib/fluent/plugin/kafka_plugin_util.rb#L35

- in module

```
def setup_kafka_client
   # patch code here
end
```

- in plugin code

```
@kafka = setup_kafka_client
```

tomiloza:
Can we merge this into master, we would really need this!

repeatedly:
@roche-d How about my comment?

repeatedly:
No reply. We need refactoring in the future but merged.

PR #175, opened 2018-01-12
removing ternary for seed_brokers, result is as reqd. on no match split

abhishekkr:
in ruby a split even without split char present will return require list

```
>> "a:4321,b:4321,c:4321".split(",")
=> ["a:4321", "b:4321", "c:4321"]
>> "a:4321".split(",")
=> ["a:4321"]
```
Signed-off-by: AbhishekKr <abhikumar163@gmail.com>


repeatedly:
Sorry for late merge.
Thanks for the patch.

Issue #112, opened 2017-02-16
Parsing broker hosts/ports from zookeeper should refer to "endpoints" section.


saturday-shi:
If some authentication framework like SASL was used in Kafka, broker hosts/ports would be registered to zookeeper in `endpoints` section. That is, left `host` and `port` with default value null and -1 ([KAFKA-3948](https://issues.apache.org/jira/browse/KAFKA-3948)). e.g.:
```
{"jmx_port":9393,"timestamp":"1487207559594","endpoints":["SASL_PLAINTEXT://host003.test.local:9092"],"host":null,"version":3,"port":-1}
```

Thus, [the latest code for now](https://github.com/fluent/fluent-plugin-kafka/blob/master/lib/fluent/plugin/out_kafka_buffered.rb#L94) will result in a list of `[":-1", ":-1", ...]`. We should look under endpoints section while parsing broker hosts/ports from zookeeper.

repeatedly:
Ah, I first knew this behaviour.
Using zookeeper is hard...

saturday-shi:
I hadn't known about this util I attempted to enable Kerberos in my cluster.
This is not a big problem for now but maybe helpful when you add something like Kerberos supporting in the future :)

PR #196, opened 2018-05-16
pickup ssl_endpoint from zookeeper node

sawanoboly:
Hi,

When certificate is specified in configuration, I would like to extract endpoint which is SSL from endpoints obtained from Zookeeper.

The version of kafka used for check is 1.0.0.

For example, when I set `PLAINTEXT: //192.0.2.11: 9092, SSL: //192.0.2.11: 9093` to `ADVERTISED_LISTENERS`. Broker information which can be retrieved from `zookeeper:///brokers/ids/1001` is as follows.

```
{
  "listener_security_protocol_map": {
    "PLAINTEXT": "PLAINTEXT",
    "SSL": "SSL"
  },
  "endpoints": [
    "PLAINTEXT://192.0.2.11:9092",
    "SSL://192.0.2.11:9093"
  ],
  "jmx_port": -1,
  "host": "192.0.2.11",
  "timestamp": "1526455616183",
  "port": 9092,
  "version": 4
}
```

### DIFF
--- a/lib/fluent/plugin/kafka_plugin_util.rb
+++ b/lib/fluent/plugin/kafka_plugin_util.rb
@@ -22,6 +22,12 @@ module Fluent
           File.read(path)
         end
       end
+
+      def pickup_ssl_endpoint(node)
+        ssl_endpoint = node['endpoints'].find {|e| e.start_with?('SSL')}
+        raise 'no SSL endpoint found on Zookeeper' unless ssl_endpoint
+        return [URI.parse(ssl_endpoint).host, URI.parse(ssl_endpoint).port].join(':')
+      end
     end
 
     module SaslSettings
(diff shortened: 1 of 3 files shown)


sawanoboly:
BTW, it may be a little related to https://github.com/fluent/fluent-plugin-kafka/issues/112. 
What do you think about the mechanism to specify the prefix to extract from endpoints by config?

repeatedly:
Could you also apply same change to `out_kafka.rb`?

repeatedly:
> What do you think about the mechanism to specify the prefix to extract from endpoints by config?

You mean if `SASL` is specified in prefix parameter, pick endpoint is limited for `SASL`, right?

sawanoboly:
> You mean if SASL is specified in prefix parameter, pick endpoint is limited for SASL, right?

Yes, for example, we may add options such as endpoint_type to extract corresponding endpoints.

repeatedly:
> Yes, for example, we may add options such as endpoint_type to extract corresponding endpoints.

It seems reasonable for me.

sawanoboly:
Thanks! 😃

PR #199, opened 2018-06-05
Add 'CentOS Installation'

wlbhiro:
I am very sorry...
(I'm sorry to have caused you trouble.)

Could you check with this?

Add CentOS Installation to README.md.

Signed-off-by: wlbhiro <wlbhiro@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,10 @@ On Ubuntu, need development packages and snappy library.
 
     $ sudo apt-get install build-essential autoconf automake libtool libsnappy-dev
 
+On CentOS 7 installation is also necessary.
+
+    $ sudo yum install gcc autoconf automake libtool snappy-devel
+
 #### Load balancing
 
 Messages will be assigned a partition at random as default by ruby-kafka, but messages with the same partition key will always be assigned to the same partition by setting `default_partition_key` in config file.


README.md
```diff
@@ -197,6 +197,10 @@ On Ubuntu, need development packages and snappy library.
 
     $ sudo apt-get install build-essential autoconf automake libtool libsnappy-dev
 
+On CentOS 7 installation is also necessary.
+
+    $ sudo yum install gcc autoconf automake libtool snappy-devel
```
cosmo0920: I guess that we can execute `yum groupinstall "Development tools"` instead of `yum install gcc autoconf automake libtool`.
How about this?

wlbhiro: Thank you for reply.

I think that "Delevelopment tools" is sufficient. 
However, I think that many unnecessary packages will be installed.

In addition, "groupinstall" may not be available when using offline local repository.
Therefore, I wrote to install only the minimum necessary package.

If "Delevelopment tools" is better, we will rewrite it.

Regard.

cosmo0920: :+1:
I see.
It looks enough for minimum installation.

wlbhiro: Thank you.
Please let me know if there are anything else.

repeatedly:
Thx!

PR #201, opened 2018-07-09
out_kafka2: fix to include event_emitter helper

t-osakada:
https://github.com/fluent/fluent-plugin-kafka/blob/master/lib/fluent/plugin/out_kafka2.rb#L132-L140

To use `router.emit()` with the v1 API output plugin, it must include `event_emitter` helper.


repeatedly:
Good catch!

Issue #192, opened 2018-04-19
Cannot auto reconnect kafka


davidwtf:
I meet a problem: fluentd will not auto reconnect kafka when kafka is temporarily offline and recovered.

I'm using kafka group input plugin.

The configs are here:
```
 <source>
      @type kafka_group
      brokers kafka1.local:9092
      topics "#{ENV['KAFKA_TOPICS']}"
      consumer_group "#{ENV['KAFKA_CONSUMER_GROUP']}"
      start_from_beginning false
      format json
      ssl_ca_cert /fluentd/etc/ca.crt
      ssl_client_cert /fluentd/etc/client.crt
      ssl_client_cert_key /fluentd/etc/client.key
      add_prefix kafka
      max_wait_time 5
 </source>
```
The logs are here:
```
2018-04-18 23:29:22 +0000 [error]: #0 unexpected error during consuming events from kafka. Re-fetch events. error="Failed to find group coordinator"
  2018-04-18 23:29:22 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/ruby-kafka-0.5.2/lib/kafka/cluster.rb:136:in `get_group_coordinator'
  2018-04-18 23:29:22 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/ruby-kafka-0.5.2/lib/kafka/consumer_group.rb:179:in `coordinator'
  2018-04-18 23:29:22 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/ruby-kafka-0.5.2/lib/kafka/consumer_group.rb:120:in `block in join_group'
  2018-04-18 23:29:22 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/ruby-kafka-0.5.2/lib/kafka/instrumenter.rb:21:in `instrument'
  2018-04-18 23:29:22 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/ruby-kafka-0.5.2/lib/kafka/instrumenter.rb:33:in `instrument'
  2018-04-18 23:29:22 +0000 [error]: #0 /usr/lib/ruby/gems/2.3.0/gems/ruby-kafka-0.5.2/lib/kafka/consumer_group.rb:119:in `join_group'
[...]

PR #204, opened 2018-08-03
Add retry_limit,disable_retry_limit to reconnect to kafka

JooyoungJeong:
Hi. 
I think retry optioin is necessary option.
I have tested it in kubernetes environment.
please check my patch . 

issue: https://github.com/fluent/fluent-plugin-kafka/issues/192

### DIFF
--- a/lib/fluent/plugin/in_kafka_group.rb
+++ b/lib/fluent/plugin/in_kafka_group.rb
@@ -192,7 +192,7 @@ class Fluent::KafkaGroupInput < Fluent::Input
   rescue =>e
     log.error "unexpected error during re-starting consumer object access", :error => e.to_s
     log.error_backtrace
-    if @retry_count <= retry_limit or disable_retry_limit == true
+    if @retry_count <= @retry_limit or disable_retry_limit
       reconnect_consumer
     end
   end
--- a/lib/fluent/plugin/in_kafka_group.rb
+++ b/lib/fluent/plugin/in_kafka_group.rb
@@ -30,6 +30,10 @@ class Fluent::KafkaGroupInput < Fluent::Input
                :desc => "Set kafka's message key to this field"
 
   config_param :retry_wait_seconds, :integer, :default => 30
+  config_param :disable_retry_limit, :bool, :default => false,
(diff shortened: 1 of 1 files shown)


repeatedly:
Hmm... these options are needed?
Put `reconnect_consumer` in `rescue` is not enough?
`retry_limit` is for preveting infinite loop in `reconnect_consumer`?

JooyoungJeong:
@repeatedly 
Thank you for your reivew. 

these options are needed?
=> The current code only retres once.  i needed retry options(retry_limit,disable_retry_limit). 
      When kafka stopped running due to some problem, I wanted fluent-plugin-kafka to retry continuously until kafka was restored.
 
These options refer to the options supported by the Buffered Output Parameters.
(https://docs.fluentd.org/v0.12/articles/output-plugin-overview)

Put reconnect_consumer in rescue is not enough?
=> 
This will result in a continuous retry, which is the same as the disable_retry_limit option.

retry_limit is for preveting infinite loop in reconnect_consumer?
=>  Yes. 
Depending on the situation, 'retry_limit' option may also be needed.

lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -175,14 +180,21 @@ def reconnect_consumer
...
+    log.warn "Could not connect to broker. retry_time:#{@retry_count}. Next retry will be in #{@retry_wait_seconds} seconds"
+    @retry_count = @retry_count + 1
     sleep @retry_wait_seconds
     @consumer = setup_consumer
     log.warn "Re-starting consumer #{Time.now.to_s}"
+    @retry_count = 0 
   rescue =>e
     log.error "unexpected error during re-starting consumer object access", :error => e.to_s
     log.error_backtrace
+    if @retry_count <= retry_limit or disable_retry_limit == true
```
repeatedly: `== true` is not needed because `disable_retry_limit` is boolean.
In addition, `@retry_limit` is better than `retry_limit`.

JooyoungJeong:
@repeatedly 
Thank you for feedback.
Please, check it again.

repeatedly:
Thanks. Merged!

PR #207, opened 2018-08-08
add parameter for exception ignore in kafka2

JiHyunSong:
hello, 
for using in output kafka2, I made it to skip ruby-kafka exception. (like skip Kafka::MessageSizeTooLarge)
and add some parameter to backup what chunks make exception occur.

I'm not used to ruby, so give me some feedbacks ^^;

thanks

### DIFF
--- a/lib/fluent/plugin/out_kafka2.rb
+++ b/lib/fluent/plugin/out_kafka2.rb
@@ -36,6 +36,9 @@ DESC
 
     config_param :get_kafka_client_log, :bool, :default => false
 
+    config_param :ignore_exceptions, :string, :default => nil, :desc => "Ignore exception list split by comma (,)"
+    config_param :exception_backup, :bool, :default => true, :desc => "chunk backup when ignore exception occured"
+
     # ruby-kafka producer options
     config_param :max_send_retries, :integer, :default => 2,
                  :desc => "Number of times to retry sending of messages to a leader."
@@ -140,6 +143,7 @@ DESC
       end
 
       @topic_key_sym = @topic_key.to_sym
+      @ignore_exceptions = if ignore_exceptions.nil?; [] else ignore_exceptions.split(',') end
     end
 
     def multi_workers_ready?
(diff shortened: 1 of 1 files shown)


soultemple:
👍 We need it to handle Kafka::MessageSizeTooLarge.

lib/fluent/plugin/out_kafka2.rb
```diff
@@ -36,6 +36,9 @@ class Fluent::Kafka2Output < Output
 
     config_param :get_kafka_client_log, :bool, :default => false
 
+    config_param :ignore_exceptions, :string, :default => nil, :desc => "Ignore exception list split by comma (,)"
```
repeatedly: use `array` type, not `string`: https://docs.fluentd.org/v1.0/articles/api-config-types#:array

lib/fluent/plugin/out_kafka2.rb
```diff
@@ -218,12 +222,22 @@ def write(chunk)
         end
       end
     rescue Exception => e
+      ignore = false
+      if ignore_exceptions.size > 0
```
repeatedly: Use `Array#include?`.

lib/fluent/plugin/out_kafka2.rb
```diff
@@ -218,12 +222,22 @@ def write(chunk)
...
+        }
+      end
+
       log.warn "Send exception occurred: #{e}"
       log.warn "Exception Backtrace : #{e.backtrace.join("\n")}"
+      log.warn "Exception ignored in tag : #{tag}" if ignore
       # For safety, refresh client and its producers
       refresh_client(false)
+      # raise UnrecoverableError for backup ignored exception chunk
+      raise Fluent::UnrecoverableError if exception_backup
```
repeatedly: `if ignore && exception_backup`?
Recoverable errors are also ignored?

JiHyunSong: `if ignore && exception_backup` (even if Fluent::UnrecoverableError defined in conf) Recoverable error will not ignored.

what do you think about which is the best way to backup chunk? (I want to know which log makes exception)

JiHyunSong:
@repeatedly check it again, please ... :D

repeatedly:
Thanks for the patch!

PR #209, opened 2018-08-14
Fix mismatch between config param and reference of message_key_key.

iwasakims:
Setting message key by message_key_key in config file does not work due to mismatch between the definition and the reference in out_kafka2.


repeatedly:
Oops. Thanks!

PR #211, opened 2018-08-22
Add support for ssl_ca_certs_from_system

danmx:
Solves https://github.com/fluent/fluent-plugin-kafka/issues/189

PR #212, opened 2018-08-22
Full SSL authentication support for input plugin

danmx:
Now input plugin supports also both SCRAM and PLAIN authentication


danmx:
Basically, there were inconsistencies between the output and input plugins. `out_kafka` supported 3 types of SSL authentication (gssapi, plain, scram; https://github.com/fluent/fluent-plugin-kafka/blob/master/lib/fluent/plugin/out_kafka.rb#L104). I just copied the same pattern to `in_kafka`.

danmx:
Resolved conflict with https://github.com/fluent/fluent-plugin-kafka/pull/211

repeatedly:
Thanks!

PR #214, opened 2018-08-28
Fixing undefined local variable or method: logger

danmx:
Fixing undefine local variable which was introduced in https://github.com/fluent/fluent-plugin-kafka/pull/212

### DIFF
--- a/lib/fluent/plugin/in_kafka.rb
+++ b/lib/fluent/plugin/in_kafka.rb
@@ -174,16 +174,16 @@ class Fluent::KafkaInput < Fluent::Input
     opt[:min_bytes] = @min_bytes if @min_bytes
 
     if @scram_mechanism != nil && @username != nil && @password != nil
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_kafka.rb
```diff
@@ -174,16 +174,16 @@ def start
     opt[:min_bytes] = @min_bytes if @min_bytes
 
     if @scram_mechanism != nil && @username != nil && @password != nil
-      @kafka = Kafka.new(seed_brokers: @brokers, client_id: @client_id, logger: logger, ssl_ca_cert: read_ssl_file(@ssl_ca_cert),
```
cosmo0920: How about use `log` instead of removing it?
`Fluent::Plugin` has `log` method.

danmx: Changed

repeatedly:
Thanks!

PR #216, opened 2018-09-11
Be tunable fetcher_max_queue_size for kafka_group

kimutansk:
### Purpose 

-  Be tunable fetcher_max_queue_size because of tuning fluentd's max process memory size.

### Why?

- Now, ruby-kafka's Fetcher holds fetched batch messages in memory.
  - https://github.com/zendesk/ruby-kafka/blob/v0.6.8/lib/kafka/fetcher.rb#L13
  - (default 100) https://github.com/zendesk/ruby-kafka/blob/v0.6.8/lib/kafka/client.rb#L291
- But, Fetcher's message means "Batched message fetched from all partisions what be assined to consumer.".
  - So, Fetcher's message size is estimated blow.
    - (One fetch result from partition , Max max_bytes=1MB) * (Num of Assined Partition) * fetcher_max_queue_size(100) * Compression Rate
      - Max 500MB over per partition.
    - Therefore, in environment what has many partitions, fluentd process's memory size has become very large.
      - 5GB or over.
- So, if fetcher_max_queue_size is tunable, more useful for tuning fluentd.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,7 @@ Consume events by kafka consumer group features..
       min_bytes               (integer) :default => nil (Use default of ruby-kafka)
       offset_commit_interval  (integer) :default => nil (Use default of ruby-kafka)
       offset_commit_threshold (integer) :default => nil (Use default of ruby-kafka)
+      fetcher_max_queue_size  (integer) :default => nil (Use default of ruby-kafka)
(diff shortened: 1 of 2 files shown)


repeatedly:
Looks good. Thanks!

PR #218, opened 2018-09-14
Support to not force SSL with SASL

Verhaeg:
Adding support to not use SSL when using SASL

According to the documentation it is possible to not force the use of SSL when using SASL authentication.
Since in some cases the Kafka cluster is inside the same network without external access, using SSL is a bit of overkill.

Added the possibility to not use SSL for the Fluentd plugin by passing the **sasl_over_ssl false**

Updated required version of ruby-kafka to 0.7.1 due to the support of not using SSL with SASL in that version

Also had to update the Producer.produce2 method (and renamed to Producer.produce_for_buffered) to comply with the new call to PendingMessage.new()

### DIFF
--- a/fluent-plugin-kafka.gemspec
+++ b/fluent-plugin-kafka.gemspec
@@ -13,12 +13,12 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.name          = "fluent-plugin-kafka"
   gem.require_paths = ["lib"]
-  gem.version       = '0.7.9'
+  gem.version       = '0.8.0'
   gem.required_ruby_version = ">= 2.1.0"
 
   gem.add_dependency "fluentd", [">= 0.10.58", "< 2"]
   gem.add_dependency 'ltsv'
-  gem.add_dependency 'ruby-kafka', '>= 0.4.1', '< 0.7.0'
+  gem.add_dependency 'ruby-kafka', '>= 0.7.1', '< 0.8.0'
   gem.add_development_dependency "rake", ">= 0.9.2"
   gem.add_development_dependency "test-unit", ">= 3.0.8"
 end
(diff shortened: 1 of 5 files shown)


Verhaeg:
Hi, any update on this? We are currently waiting this to be able to use in production.

Thanks

fluent-plugin-kafka.gemspec
```diff
@@ -13,12 +13,12 @@ Gem::Specification.new do |gem|
   gem.test_files    = gem.files.grep(%r{^(test|spec|features)/})
   gem.name          = "fluent-plugin-kafka"
   gem.require_paths = ["lib"]
-  gem.version       = '0.7.9'
+  gem.version       = '0.8.0'
```
repeatedly: Could you remove this change?

lib/fluent/plugin/kafka_producer_ext.rb
```diff
@@ -10,16 +10,17 @@
 # for out_kafka_buffered
 module Kafka
   class Producer
-    def produce2(value, key: nil, topic:, partition: nil, partition_key: nil)
+    def produce_for_buffered(value, key: nil, topic:, partition: nil, partition_key: nil)
       create_time = Time.now
 
       message = PendingMessage.new(
```
repeatedly: Should apply same changes to `TopicProducer#produce`?

Verhaeg: I did change there, but not quite sure if should use produce() or produce_for_buffered() there.
For now using produce_for_buffered, but if I'm mistaken, please do tell me and I'll correct it asap ;)

repeatedly: I mean `out_kafka2` uses specific producer named `TopicProducer` and it also uses `PendingMessage` in its method: https://github.com/fluent/fluent-plugin-kafka/blob/91da31930601ad1a60f1d3bf9b32f3a509341ee3/lib/fluent/plugin/kafka_producer_ext.rb#L84

Verhaeg: Thanks for pointing it out ;)

repeatedly:
Thanks!

PR #224, opened 2018-11-12
 Adding support to not use SSL when using SASL on "in" plugin...

mcarbonneaux:
like "out" plugin commit 8ea31d00d4d8ba0598171e7f6dbb4be9e05724b2, this pull request add support to not use SSL when using SASL to "in" plugin....

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,12 +41,25 @@ If you want to use zookeeper related parameters, you also need to install zookee
 
 Set path to SSL related files. See [Encryption and Authentication using SSL](https://github.com/zendesk/ruby-kafka#encryption-and-authentication-using-ssl) for more detail.
 
-#### SASL authentication
+#### SASL authentication 
+
+##### with GSSAPI
 
 - principal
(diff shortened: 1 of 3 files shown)


repeatedly:
Thanks!

abernardino:
Are there plans to support this on the "out" plugin as well?

repeatedly:
> Are there plans to support this on the "out" plugin as well?

See PR description. out plugins support this parameter...

repeatedly:
> Why is it not honored for the PLAIN or other mechanism? Any plan to support that?

Because first patch covers only SASL SCRAM.
Patches for adding more mechanizm are welcome.

doublening:
Thanks. Is there document for building fluent-plugin-kafka gem locally?

mcarbonneaux:
Honoring plain text sasl over non crypted layer is risky... 

Scram can be done... 

I think is why the original author have not done for plain text...

PR #244, opened 2019-02-22
Allow setting to not use ssl with plain password

Verhaeg:
Following the same applications of sasl ha256, passing variable
sasl_over_ssl to the constructor when using plain auth.


repeatedly:
Thx!

PR #249, opened 2019-03-26
Enable the setting of partition key

KeiichiHirobe:
Enable the setting of partition_key which has been invalid due to typo.
Additionally, I found some mistakes at the section [load-balancing](https://github.com/KeiichiHirobe/fluent-plugin-kafka#load-balancing),  and I fix that.


repeatedly:
Oops. Good catch. Thanks!

PR #256, opened 2019-05-02
Expose connection timeouts

arturoera:
We were running on a problem where we ultimately solved by exposing 2 parameters from ruby-kafka connections to kafka-group ingestion: connect_timeout and socket_timeout
From here: https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/connection.rb#L49

Our remote brokers over SSL needed the socket_timeout increased from the default in order to wait for "join_group", otherwise we ended up in a non-stop rebalancing cycle at the brokers.

### DIFF
--- a/lib/fluent/plugin/in_kafka_group.rb
+++ b/lib/fluent/plugin/in_kafka_group.rb
@@ -30,9 +30,9 @@ class Fluent::KafkaGroupInput < Fluent::Input
                :desc => "Time format to be used to parse 'time' filed."
   config_param :kafka_message_key, :string, :default => nil,
                :desc => "Set kafka's message key to this field"
-  config_param :connect_timeout, :integer, :default => 10,
+  config_param :connect_timeout, :integer, :default => nil,
                :desc => "[Integer, nil] the timeout setting for connecting to brokers"
-  config_param :socket_timeout, :integer, :default => 10,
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -30,6 +30,10 @@ class Fluent::KafkaGroupInput < Fluent::Input
                :desc => "Time format to be used to parse 'time' filed."
   config_param :kafka_message_key, :string, :default => nil,
                :desc => "Set kafka's message key to this field"
+  config_param :connect_timeout, :integer, :default => 10,
```
repeatedly: default should be `nil` because ruby-kafka uses nil for default value.

lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -30,6 +30,10 @@ class Fluent::KafkaGroupInput < Fluent::Input
...
+               :desc => "[Integer, nil] the timeout setting for connecting to brokers"
+  config_param :socket_timeout, :integer, :default => 10,
```
repeatedly: ditto.

arturoera: changed to the default nil values

repeatedly:
Thanks!

PR #259, opened 2019-05-14
`chunk.metadata.variables` could be nil

ganmacs:
Without this patch, No method error could be raised here.

@repeatedly Could you review this?


repeatedly:
Seems good.

Issue #255, opened 2019-04-30
challenges configuring Kerberos without SSL


AdimUser:
I am planning to send logs to our cooperate elastic search cluster.

I am using td-agent to forward logs to kafka endpoints which is connected to cooperate elastic search cluster. I have installed the fluent-plugin-kafka plugin with td-agent to send logs to kafka endpoints which is connected to cooperate elastic search cluster

LOGS SENDING PATHS AS BELOW:

`
logs --> td-agent+fluent-plugin-kafka --> kafka cluster --> elasticsearch --> kibana
`

Issue is, kafka endpoints are configured in a way to support SASL AUTHENTICATION without supporting SSL, Since logs are only route with cooperate network, I guess no need of have SSL support.

I am getting bellow error :
```
2019-04-30 17:46:39 +0900 [error]: #0 /opt/td-agent/embedded/lib/ruby/gems/2.4.0/gems/fluentd-1.3.3/bin/fluentd:8:in `<top (required)>'
  2019-04-30 17:46:39 +0900 [error]: #0 /opt/td-agent/embedded/bin/fluentd:23:in `load'
  2019-04-30 17:46:39 +0900 [error]: #0 /opt/td-agent/embedded/bin/fluentd:23:in `<main>'
2019-04-30 17:46:39 +0900 [error]: #0 unexpected error error_class=ArgumentError error="SASL authentication requires that SSL is configured"
  2019-04-30 17:46:39 +0900 [error]: #0 suppressed same stacktrace
2019-04-30 17:46:39 +0900 [info]: Worker 0 finished unexpectedly with status 1
^C
```

My td-agent configs are :

```
<source>
  @type dummy
  dummy {"hello":"world"}
  tag test
</source>

<match test>
  @type kafka2
  brokers               <broker1>:9002,<broker2>:9002,<<broker3>>:9002
  principal             'appuser@STGKAFKA100.GEAP.XXX.COM'
  keytab                'appuser.keytab'
  client_id             'kafka'
  sasl_over_ssl         false
  get_kafka_client_log  true
  <format>
    @type json
  </format>
  topic_key             'stg_app_elk_1'
  get_kafka_client_log  true

  <buffer topic>
    flush_interval 10s
  </buffer>
</match>
```

repeatedly:
[...]

PR #261, opened 2019-05-22
Added setting to not use SSL with SASL/GSSAPI for Kerberos Authentication

mihir2402:
Similar to #244 but for Kerberos Authentication.

Related to issue  #255


repeatedly:
Thanks! Release new version soon.

PR #262, opened 2019-05-24
Update links to docs.fluentd.org

okkez:


cosmo0920:
Looks reasonable for me.
But I have a nitpicks comment.

Should we use `https://docs.fluentd.org/v/1.0/` instead of `https://docs.fluentd.org/` for document URL prefixes which are pointed v1.0 version clearly?

repeatedly:
Hmm... add prefix can avoid confusion.

repeatedly:
Thanks!

Issue #263, opened 2019-05-28
kafka2 - Exception on write()


kevinkirkup:
When attempting to write a chunk using the `kafka2` output plugin, I'm getting the following exception:

```
2019-05-28 19:53:46 +0000 [warn]: #0 Send exception occurred: undefined method `[]' for nil:NilClass
2019-05-28 19:53:46 +0000 [warn]: #0 Exception Backtrace : /opt/td-agent/embedded/lib/ruby/gems/2.4.0/gems/fluent-plugin-kafka-0.9.4/lib/fluent/plugin/out_kafka2.rb:206:in `write'
/opt/td-agent/embedded/lib/ruby/gems/2.4.0/gems/fluentd-1.4.2/lib/fluent/plugin/output.rb:1125:in `try_flush'
/opt/td-agent/embedded/lib/ruby/gems/2.4.0/gems/fluentd-1.4.2/lib/fluent/plugin/output.rb:1425:in `flush_thread_run'
/opt/td-agent/embedded/lib/ruby/gems/2.4.0/gems/fluentd-1.4.2/lib/fluent/plugin/output.rb:454:in `block (2 levels) in start'
/opt/td-agent/embedded/lib/ruby/gems/2.4.0/gems/fluentd-1.4.2/lib/fluent/plugin_helper/thread.rb:78:in `block in thread_create'
2019-05-28 19:53:46 +0000 [info]: #0 initialized kafka producer: fluentd
2019-05-28 19:53:46 +0000 [warn]: #0 got unrecoverable error in primary and no secondary error_class=NoMethodError error="undefined method `[]' for nil:NilClass"
  2019-05-28 19:53:46 +0000 [warn]: #0 suppressed same stacktrace
2019-05-28 19:53:46 +0000 [warn]: #0 bad chunk is moved to /tmp/fluent/backup/worker0/object_2afda776ef10/589f80583e7651ee41a4734a66535ade.log
```

Looks like the code assumes that `chunk.metadata.variables` is always initialized, but the default is `nil`. Need to add handling for that.

repeatedly:
Do you use v0.12 chunk files with kafka2?

kevinkirkup:
No, I'm using v1.0.

Looks like this was fixed for the `rdkafka2`
https://github.com/fluent/fluent-plugin-kafka/blob/c6ec4331c7f16b2800804f120e8a7ed8409d61aa/lib/fluent/plugin/out_rdkafka2.rb#L233

[...]

PR #265, opened 2019-05-28
Fix chunk.metadata.variables can be nil for kafka2

kevinkirkup:
`chunk.metadata.varialbes` can be and defaults to `nil`.
Need to handle that case.

This was already fixed for rdkafka2.

Signed-off-by: Kevin S Kirkup <kevin.kirkup@pureport.com>


kevinkirkup:
Fixes #263

repeatedly:
Thanks. I forgot to fix this in rdkafka2...

ganmacs:
Thanks a lot.

PR #270, opened 2019-06-07
add support for ouput kafka with zstd compression

like-inspur:
I also found ruby-kafka support zstd compression on newest release, I test it and it works!


repeatedly:
Thx!

PR #272, opened 2019-06-19
Add configuration for ssl_verify_hostname

ogibayashi:
Hi. I would like to configure `ssl_verify_hostname` of ruby-kafka which is added in https://github.com/zendesk/ruby-kafka/pull/733

### DIFF
--- a/fluent-plugin-kafka.gemspec
+++ b/fluent-plugin-kafka.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "fluentd", [">= 0.10.58", "< 2"]
   gem.add_dependency 'ltsv'
-  gem.add_dependency 'ruby-kafka', '>= 0.7.1', '< 0.8.0'
+  gem.add_dependency 'ruby-kafka', '>= 0.7.8', '< 0.8.0'
   gem.add_development_dependency "rake", ">= 0.9.2"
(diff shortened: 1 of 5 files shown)


repeatedly:
Thanks. I will check after back from kubecon.

fluent-plugin-kafka.gemspec
```diff
@@ -18,7 +18,7 @@ Gem::Specification.new do |gem|
 
   gem.add_dependency "fluentd", [">= 0.10.58", "< 2"]
   gem.add_dependency 'ltsv'
-  gem.add_dependency 'ruby-kafka', '>= 0.7.1', '< 0.8.0'
+  gem.add_dependency 'ruby-kafka', '>= 0.7.8', '< 0.8.0'
```
ogibayashi: `ssl_verify_hostname` is available with ruby-kafka >= 0.7.8.

ganmacs:
Doesn't it need to add the same change as out_kafka2.rb to out_kafka.rb?

ogibayashi:
That's right. I have added the change to out_kafka2.rb and out_kafka_buffered.rb.

ganmacs:
I noticed now but are in_kafka.rb and in_kafka_group.rb also needed?

repeatedly:
I think adding it better but not required for this patch.

repeatedly:
Thanks!

ogibayashi:
Thank you!

PR #277, opened 2019-07-03
Enable defining headers for kafka2 and rdkafka2 output

PhilGrayson:
I'd like to be able to define headers on a Kafka message based on the content of a fluentd record.

This is due to the Splunk Kafka Connect plugin using Kafka headers to populate some metadata about an event in Splunk ([source](https://github.com/splunk/kafka-connect-splunk#headers-parameters)).

---

I've raised this PR to see if the maintainer is interested in the contribution. No offence taken if not, I'd be wary about accepting new code without any automated tests.

### DIFF
--- a/lib/fluent/plugin/out_kafka2.rb
+++ b/lib/fluent/plugin/out_kafka2.rb
@@ -39,8 +39,10 @@ DESC
                  :desc => 'Set true to remove partition key from data'
     config_param :exclude_topic_key, :bool, :default => false,
                  :desc => 'Set true to remove topic name key from data'
-    config_param :headers, :hash, default: {}, symbolize_keys: true, value_type: :string
-    config_param :headers_from_record, :hash, default: {}, symbolize_keys: true, value_type: :string
+    config_param :headers, :hash, default: {}, symbolize_keys: true, value_type: :string,
+                 :desc => 'Kafka message headers'
+    config_param :headers_from_record, :hash, default: {}, symbolize_keys: true, value_type: :string,
+                 :desc => 'Kafka message headers where the header value is a jsonpath to a record value'
 
     config_param :get_kafka_client_log, :bool, :default => false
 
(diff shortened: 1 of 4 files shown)


PhilGrayson:
Headers should be an array, not a Hash. Kafka allows a record to have headers with duplicate keys.

lib/fluent/plugin/out_kafka2.rb
```diff
@@ -39,6 +39,8 @@ class Fluent::Kafka2Output < Output
                  :desc => 'Set true to remove partition key from data'
     config_param :exclude_topic_key, :bool, :default => false,
                  :desc => 'Set true to remove topic name key from data'
+    config_param :headers, :hash, default: {}, symbolize_keys: true, value_type: :string
```
repeatedly: need `:desc` paramaeter.

lib/fluent/plugin/out_kafka2.rb
```diff
@@ -39,6 +39,8 @@ class Fluent::Kafka2Output < Output
...
+    config_param :headers_from_record, :hash, default: {}, symbolize_keys: true, value_type: :string
```
repeatedly: ditto

PhilGrayson:
Thanks for the review! I've pushed a new commit.

I've considered adding this feature to the 'kafka' and 'kafka_buffered' output. However it'll require a bump to the minimum fluentd version dependency. Do you have any thoughts on this?

Current gemspec has
```
```
The handy 'record_accessor' helper was added in fluentd v0.14.20

PhilGrayson:
> Headers should be an array, not a Hash. Kafka allows a record to have headers with duplicate keys.

Turns out [ruby-kafka](https://github.com/zendesk/ruby-kafka/blob/dce5db4d6ef7a8ec6dd0dc231df3a4029ecde05b/lib/kafka/producer.rb#L182) and [rdafka-ruby](https://github.com/appsignal/rdkafka-ruby/blob/10fe675583a0f33866357e79ddaf83025bb1f566/lib/rdkafka/producer.rb#L60]) accepts headers as a Hash only.

repeatedly:
> I've considered adding this feature to the 'kafka' and 'kafka_buffered' output. However it'll require a bump to the minimum fluentd version dependency. Do you have any thoughts on this?

I think no need support for old API plugins. Copy `record_accessor` code is overenginnering for me.

repeatedly:
Thanks!

Issue #273, opened 2019-06-19
How to specify the time key?


rchady:
Running in to an issue where messages sent to Kafka from Fluentd have message.create_time = -1.  Is there a config option I need to set to specify what to use for the create_time?

repeatedly:
No need to specify because producer always sets current time to message.create_time.

https://github.com/fluent/fluent-plugin-kafka/blob/740c8fb3bac025b29b7e69ea3807fbacf78a5258/lib/fluent/plugin/kafka_producer_ext.rb#L16
https://github.com/fluent/fluent-plugin-kafka/blob/740c8fb3bac025b29b7e69ea3807fbacf78a5258/lib/fluent/plugin/kafka_producer_ext.rb#L103

rchady:
Is there a way to specify what it should be?  For example, I have a key inside my data I'd like to use for the time vs. when Fluentd gets the message.

Also, I'm not seeing it?  When I inspect a message from Kafka, I'm seeing:

```
(byebug) message.create_time
1969-12-31 17:59:59 -0600
```

repeatedly:
> For example, I have a key inside my data I'd like to use for the time vs. when Fluentd gets the message.

Basically, if your data has a time, it should be fluentd event time, not record field. This is done by parser or filter.
To use event time instead of current time, the patch is needed like below:

```
diff --git a/lib/fluent/plugin/kafka_producer_ext.rb b/lib/fluent/plugin/kafka_producer_ext.rb
index 486dde5..f13134f 100644
--- a/lib/fluent/plugin/kafka_producer_ext.rb
+++ b/lib/fluent/plugin/kafka_producer_ext.rb
@@ -12,9 +12,7 @@ module Kafka
   EMPTY_HEADER = {}
 
   class Producer
-    def produce_for_buffered(value, key: nil, topic:, partition: nil, partition_key: nil)
-      create_time = Time.now
-
+    def produce_for_buffered(value, key: nil, topic:, partition: nil, partition_key: nil, create_time: Time.now)
       message = PendingMessage.new(
         value: value,
         key: key,
@@ -99,9 +97,7 @@ module Kafka
       @pending_message_queue = PendingMessageQueue.new
     end
 
-    def produce(value, key: nil, partition: nil, partition_key: nil)
-      create_time = Time.now
-
+    def produce(value, key: nil, partition: nil, partition_key: nil, create_time: Time.now)
       message = PendingMessage.new(
         value: value,
         key: key,
[...]

PR #278, opened 2019-07-05
Support specifying CreateTime from a record field

j30ng:
Hello. I ran into a use case where I need to set CreateTIme from a record field. I added a config_param `create_time_key` so that users can specify a field containing the time data they want to use as CreateTime. Please consider this PR, and let me know if you approve of it!

### DIFF
--- a/lib/fluent/plugin/out_kafka2.rb
+++ b/lib/fluent/plugin/out_kafka2.rb
@@ -20,6 +20,7 @@ DESC
                  :desc => "Default output topic when record doesn't have topic field"
     config_param :message_key_key, :string, :default => 'message_key', :desc => "Field for kafka message key"
     config_param :default_message_key, :string, :default => nil
+    config_param :create_time_key, :string, :default => nil, :desc => "Field for CreateTime"
     config_param :partition_key_key, :string, :default => 'partition_key', :desc => "Field for kafka partition key"
     config_param :default_partition_key, :string, :default => nil
     config_param :partition_key, :string, :default => 'partition', :desc => "Field for kafka partition"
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_kafka2.rb
```diff
@@ -229,7 +230,9 @@ def write(chunk)
           log.trace { "message will send to #{topic} with partition_key: #{partition_key}, partition: #{partition}, message_key: #{message_key} and value: #{record_buf}." }
           messages += 1
 
-          producer.produce(record_buf, key: message_key, partition_key: partition_key, partition: partition)
+          create_time = @create_time_key != nil ? record[@create_time_key]: nil
```
ganmacs: [nits] `@create_time_key && record[@create_time_key]`

lib/fluent/plugin/out_kafka_buffered.rb
```diff
@@ -339,7 +340,10 @@ def write(chunk)
         end
         log.trace { "message will send to #{topic} with partition_key: #{partition_key}, partition: #{partition}, message_key: #{message_key} and value: #{record_buf}." }
         messages += 1
-        producer.produce_for_buffered(record_buf, topic: topic, key: message_key, partition_key: partition_key, partition: partition)
+
+        create_time = @create_time_key != nil ? record[@create_time_key]: nil
```
ganmacs: ditto.

repeatedly:
Maybe, this approach is not correct. Fluentd event has time so we can use it instead of field value.

https://github.com/fluent/fluent-plugin-kafka/issues/273#issuecomment-504099585

I wrote a draft for this feature.

j30ng:
> Maybe, this approach is not good. Fluentd event has time so we can use it instead of field value. If you need to use field value for the time, it should be parsed before this plugin.

That is a nicer approach, you're right. I'm fairly new to Fluentd, and I didn't know you could do that. Your patch by the way looks quite complete. Do you have any plans to merge it?

repeatedly:
I just created new PR with above. https://github.com/fluent/fluent-plugin-kafka/pull/279
I want someone tests this patch.

repeatedly:
#279 merged.

PR #283, opened 2019-08-06
Add option to use Kafka's timestamps as the time source

stevenjm:
This replaces the `use_record_time` boolean option with a `time_source` enum option to select the desired time source, with `time_source kafka` providing timestamps from Kafka itself. `use_record_time true` is now a deprecated setting equivalent to `time_source record`.

This also fixes `time_source record` in the `in_kafka` plugin (previously, `use_record_time` did nothing with `in_kafka`, it only worked with `in_kafka_group`).

Note that in order to behave correctly, `time_source kafka` requires a version of ruby-kafka with the fix in https://github.com/zendesk/ruby-kafka/pull/766.


repeatedly:
Ahhhhhhhhhh, sorry.
I forgot to click `submit` button so my reviews are pending over a month... ;(

stevenjm:
@repeatedly No problem, thanks for the review!

Addressed your comments and rebased onto master.

repeatedly:
Thanks for the patch. New version will be released soon.

PR #303, opened 2019-12-05
Add sasl_over_ssl parameter to plain sasl consumer constructor calls

StevTheDev:
This change adds the sasl_over_ssl paramter to plain sasl consumer
constructor calls so that sasl_over_ssl can be disabled.

Signed-off-by: Steven Fitzpatrick <StevTheDev@gmail.com>

### DIFF
--- a/lib/fluent/plugin/in_kafka.rb
+++ b/lib/fluent/plugin/in_kafka.rb
@@ -190,7 +190,8 @@ class Fluent::KafkaInput < Fluent::Input
     elsif @username != nil && @password != nil
       @kafka = Kafka.new(seed_brokers: @brokers, client_id: @client_id, logger: logger, ssl_ca_cert: read_ssl_file(@ssl_ca_cert),
                          ssl_client_cert: read_ssl_file(@ssl_client_cert), ssl_client_cert_key: read_ssl_file(@ssl_client_cert_key),
-                         ssl_ca_certs_from_system: @ssl_ca_certs_from_system,sasl_plain_username: @username, sasl_plain_password: @password)
+                         ssl_ca_certs_from_system: @ssl_ca_certs_from_system,sasl_plain_username: @username, sasl_plain_password: @password,
+                         sasl_over_ssl: @sasl_over_ssl)
     else
       @kafka = Kafka.new(seed_brokers: @brokers, client_id: @client_id, logger: logger, ssl_ca_cert: read_ssl_file(@ssl_ca_cert),
                          ssl_client_cert: read_ssl_file(@ssl_client_cert), ssl_client_cert_key: read_ssl_file(@ssl_client_cert_key),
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_kafka.rb
```diff
@@ -190,7 +190,8 @@ def start
     elsif @username != nil && @password != nil
       @kafka = Kafka.new(seed_brokers: @brokers, client_id: @client_id, logger: logger, ssl_ca_cert: read_ssl_file(@ssl_ca_cert),
                          ssl_client_cert: read_ssl_file(@ssl_client_cert), ssl_client_cert_key: read_ssl_file(@ssl_client_cert_key),
-                         ssl_ca_certs_from_system: @ssl_ca_certs_from_system,sasl_plain_username: @username, sasl_plain_password: @password)
+                         ssl_ca_certs_from_system: @ssl_ca_certs_from_system,sasl_plain_username: @username, sasl_plain_password: @password,
+                         sasl_over_ssl: @sasl_over_ssl)
```
ganmacs: I'm not sure but adding the same code at https://github.com/fluent/fluent-plugin-kafka/pull/303/files#diff-16a11d1357395c8e8484ae617f831940R196 is better?

lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -170,7 +170,8 @@ def start
     elsif @username != nil && @password != nil
       @kafka = Kafka.new(seed_brokers: @brokers, client_id: @client_id, logger: logger, connect_timeout: @connect_timeout, socket_timeout: @socket_timeout, ssl_ca_cert: read_ssl_file(@ssl_ca_cert),
                          ssl_client_cert: read_ssl_file(@ssl_client_cert), ssl_client_cert_key: read_ssl_file(@ssl_client_cert_key),
-                         ssl_ca_certs_from_system: @ssl_ca_certs_from_system, sasl_plain_username: @username, sasl_plain_password: @password)
+                         ssl_ca_certs_from_system: @ssl_ca_certs_from_system, sasl_plain_username: @username, sasl_plain_password: @password,
+                         sasl_over_ssl: @sasl_over_ssl)
```
ganmacs: ditto https://github.com/fluent/fluent-plugin-kafka/pull/303/files#diff-a3e302d26304c1507f4b5f008652dffeR176

lib/fluent/plugin/in_kafka.rb
StevTheDev: I don't think so. In that case neither `scram_mechanism`, `username` or `password` are provided. If I am reading [this](https://github.com/zendesk/ruby-kafka#authentication-using-sasl) correctly, omitting these parameters means that no SASL auth will be attempted so adding `sasl_over_ssl` to that case is not necessary.

lib/fluent/plugin/in_kafka_group.rb
StevTheDev: I have the same thoughts here as above

lib/fluent/plugin/in_kafka.rb
ganmacs: Make sense.

StevTheDev:
@ganmacs  as an aside - if this change gets merged in, do you know when I could expect the next release to be made?

Thanks for reviewing!

StevTheDev:
Happy new year @repeatedly! Would you please review this change?

repeatedly:
Ah, sorry. I though this was merged...
I will release new version soon after back to home.

PR #324, opened 2020-03-02
out_rdkafka2: Follow rdkafka's log level for rdkafka log

repeatedly:
Signed-off-by: Masahiro Nakagawa <repeatedly@gmail.com>

### DIFF
--- a/lib/fluent/plugin/out_rdkafka.rb
+++ b/lib/fluent/plugin/out_rdkafka.rb
@@ -94,7 +94,7 @@ DESC
       def add(level, &block)
         return unless block
 
-        # Follow rdkakfa's log level. See also rdkafka-ruby's bindings.rb.
+        # Follow rdkakfa's log level. See also rdkafka-ruby's bindings.rb: https://github.com/appsignal/rdkafka-ruby/blob/e5c7261e3f2637554a5c12b924be297d7dca1328/lib/rdkafka/bindings.rb#L117
         case level
         when Logger::FATAL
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_rdkafka.rb
```diff
@@ -91,8 +92,22 @@ def configure(conf)
     super
     log.instance_eval {
       def add(level, &block)
-        if block
+        return unless block
+
+        # Follow rdkakfa's log level. See also rdkafka-ruby's bindings.rb.
```
ganmacs: I think it's better to have the permanent link which has commit hash to bindings.rb.

repeatedly: Added github link

lib/fluent/plugin/out_rdkafka.rb
```diff
@@ -91,8 +92,22 @@ def configure(conf)
     super
     log.instance_eval {
       def add(level, &block)
-        if block
+        return unless block
```
ganmacs: [nits] Can we use `block_given?`?

repeatedly: Does it mean change method args like below?

```
def add(level)
  return unless block_given?
```

ganmacs: I thought just changing to `return unless block_given?` part. but yours also is fine for me.

PR #338, opened 2020-05-26
Add description of time_source in README.md

JaeGeunBang:
Add description of time_source in input plugin. 
I needed kafka timestamp, but there was no description in README.md. But, for the kafka timestamp, the time_source record was already implemented in the code, and it was necessary to explain it.


repeatedly:
Ah, good catch. Thanks!

PR #342, opened 2020-06-04
Add custom time key when time_source is record

gigkokman:
When we set time_source to record, we can specify time field instead of fixed value, 'time'.


lib/fluent/plugin/in_kafka.rb
```diff
@@ -114,8 +116,10 @@ def configure(conf)
     @parser_proc = setup_parser
 
     @time_source = :record if @use_record_time
+    @time_key = 'time'
```
repeatedly, line 119: Why separate instance variable into `@time_key` and `@record_time_key`?
Does using `@record_time_key` for `record[]` have a problem?

gigkokman: You have a good point, I will update it.

repeatedly:
Thanks!

PR #358, opened 2020-08-06
Fix partitioner call in Kafka::Producer due to ruby-kafka change

sabriel27:
The newest version of ruby-kafka changed the Kafka::Partitioner API from `partition_for_key` to `call`: https://github.com/zendesk/ruby-kafka/commit/8dc495f471312834e477cd16144a32295c6b0a65#diff-0bc0466ba80bcb74058c1e4def09bf41R22


repeatedly:
Thanks!

PR #361, opened 2020-08-10
Fix for partitioner call method

dmilde:
Solves error "undefined call method" of partitioner.
Working, but I am not a ruby programmer/expert, check needed.

### DIFF
--- a/lib/fluent/plugin/kafka_producer_ext.rb
+++ b/lib/fluent/plugin/kafka_producer_ext.rb
@@ -245,13 +245,13 @@ module Kafka
 
     def assign_partitions!
       failed_messages = []
+      partition_count = @cluster.partitions_for(@topic).count
 
       @pending_message_queue.each do |message|
         partition = message.partition
 
         begin
           if partition.nil?
-            partition_count = @cluster.partitions_for(@topic).count
             partition = @partitioner.call(partition_count, message)
           end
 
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/kafka_producer_ext.rb
```diff
@@ -243,14 +245,14 @@ def deliver_messages_with_retries
...
       failed_messages = []
-      partition_count = @cluster.partitions_for(@topic).count
 
       @pending_message_queue.each do |message|
         partition = message.partition
 
         begin
           if partition.nil?
-            partition = Partitioner.call(partition_count, message)
+            partition_count = @cluster.partitions_for(@topic).count
```
repeatedly, line 254: Why move this line?
In `TopicProducer`, `@topic` is fixed value so no need to call for each message.

repeatedly:
Thanks for the patch. I put one comment.

dmilde:
Moved the count back (out of the loop).

lib/fluent/plugin/kafka_producer_ext.rb
dmilde: Move the count back (out of the loop).

repeatedly:
I confirmed this patch fixes the issue. I will release new version soon. Thanks!

PR #369, opened 2020-09-17
Fixed a small bug in the way an exception is raised.

chrisw-instaclustr:
The wrong type was being used when raising this error.


repeatedly:
Ah, good catch!

PR #379, opened 2020-11-19
Use NumericTimeParser to parse float time

underluns:
The [fluentd documentation](https://docs.fluentd.org/configuration/format-section#time-parameters) says:
```
time_type (enum) (optional): parses/formats value according to this type
Default: float (econds from Epoch + nano seconds (e.g.1510544836.154709804))
```
But the `float` type is not supported by default in this library. As a result, an [error raises](https://github.com/fluent/fluentd/blob/v1.11.5/lib/fluent/plugin/output.rb#L828):
```
time must be a Fluent::EventTime (or Integer): Float
```
I suggest using parser [Fluent::NumericTimeParser](https://github.com/fluent/fluentd/blob/v1.11.5/lib/fluent/time.rb#L282) for float times like other [plugins do](https://github.com/fluent/fluentd/blob/v1.11.5/lib/fluent/time.rb#L182).

P.S. Use [time_format "%s.%L"](https://github.com/fluent/fluent-plugin-kafka/pull/238) does not seem very correct to me.

### DIFF
--- a/lib/fluent/plugin/in_kafka_group.rb
+++ b/lib/fluent/plugin/in_kafka_group.rb
@@ -138,6 +138,10 @@ class Fluent::KafkaGroupInput < Fluent::Input
         @time_parser = Fluent::TextParser::TimeParser.new(@time_format)
       end
     end
+
+    if @time_source == :record && defined?(Fluent::NumericTimeParser)
+      @float_numeric_parse = Fluent::NumericTimeParser.new(:float)
+    end
   end
 
   def setup_parser(conf)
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -264,6 +264,8 @@ def run
               when :record
                 if @time_format
                   record_time = @time_parser.parse(record[@record_time_key].to_s)
+                elsif record[@record_time_key].is_a?(Float)
+                  record_time = Fluent::NumericTimeParser.new(:float).parse(record[@record_time_key])
```
repeatedly, line 268: Create NumberTimeParser in each record is extremely slow.
The parser should be created in `configure`.
In addition, call `record[@record_time_key]` twice is redundant. Following code is better

```
if @time_format
  # same as
else
  record_time = record[@record_time_key]
  if record_time.is_a?(Float)
    record_time = @float_numeric_parser.parse(record_time)
  end
end
```

cosmo0920 (commented):
This change seems reasonable.
@repeatedly What is the blocker to accept this patch?

repeatedly:
I forgot to merge...
New version will be released soon.

PR #381, opened 2020-12-07
Fix: record_time_key not working in in_kafka.rb

### LINKED ISSUE #380: record_time_key is not working
We noticed that setting the `time_source` to `record` in the `kafka` input plugin does not work.

It correctly uses the configured time source, however, the config_parameter `record_time_key` is not being propagated properly into the `TopicWatcher` class resulting the `record_time` being `nil`.

```
emit transaction failed: error_class=TypeError error="can't convert nil into an exact number" location="/usr/local/bundle/gems/fluentd-1.11.5/lib/fluent/plugin/formatter_stdout.rb:45:in `at'" tag="mytag"
```

When I set `record_time = record["time"]` instead of `record_time = record[@record_time_key]` in `in_kafka.rb#L308` it works. However, `@record_time_key` is just empty at this point in the code.

A quick fix is to simply add `record_time_key` as an additional argument to the class's constructor and setting the property again from it. However, as I'm not really versed in Ruby, I'm not sure if this is the best way to handle it.

Best regards
Marcel

marratj:
This PR is to fix #380 

It simply supplies `record_time_key` as additional argument in the `TopicWatcher` class constructor, which was missing until now, which led the option to use the `record` time_source to not work.

### DIFF
--- a/lib/fluent/plugin/in_kafka.rb
+++ b/lib/fluent/plugin/in_kafka.rb
@@ -224,6 +224,7 @@ class Fluent::KafkaInput < Fluent::Input
         router,
         @kafka_message_key,
         @time_source,
+        @record_time_key,
         opt)
     }
     @topic_watchers.each {|tw|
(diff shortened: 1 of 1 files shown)


repeatedly:
Thanks!

PR #386, opened 2021-01-13
Add option to use record as source for the Fluentd event tag

marratj:
This PR adds the option to set the Fluentd event tag to a record field instead of the topic name only (similar to the existing `time_source` option).

We created this on our side as we have the need to keep the source event tags when we have a Fluentd on a source cluster and another one on a target cluster where a Kafka topic is acting as queue in between.

### DIFF
--- a/lib/fluent/plugin/in_kafka.rb
+++ b/lib/fluent/plugin/in_kafka.rb
@@ -31,6 +31,8 @@ class Fluent::KafkaInput < Fluent::Input
   config_param :add_suffix, :string, :default => nil,
                :desc => "tag suffix"
   config_param :add_offset_in_record, :bool, :default => false
+  config_param :tag_source, :enum, :list => [:topic, :record], :default => :topic,
+               :desc => "Source for the fluentd event tag"
 
   config_param :offset_zookeeper, :string, :default => nil
   config_param :offset_zk_root_node, :string, :default => '/fluent-plugin-kafka'
@@ -225,6 +227,7 @@ class Fluent::KafkaInput < Fluent::Input
         @kafka_message_key,
         @time_source,
         @record_time_key,
+        @tag_source,
         opt)
     }
     @topic_watchers.each {|tw|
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -248,14 +252,18 @@ def run
     while @consumer
       begin
         @consumer.each_batch(@fetch_opts) { |batch|
-          es = Fluent::MultiEventStream.new
```
repeatedly, line 251: For the performance, separating method is better for `in_kafka_group`.

```
@consumer.each_batch(@fetch_opts) { |batch|
  if @tag_source == :record
    process_batch_with_record_tag(batch) # use this patch implementation
  else
    process_batch(batch)  # use existing implementation
  end
}
```

marratj: Thanks for the suggestion, implemented it. Care to review again? :)

marratj:
@repeatedly would you like to review again, I have implemented your suggested change :)

lib/fluent/plugin/in_kafka.rb
```diff
@@ -298,6 +306,11 @@ def consume
       messages.each { |msg|
         begin
           record = @parser.call(msg, @topic_entry)
+          if @tag_source == :record
+            tag = record[@record_tag_key]
```
cosmo0920, line 310: How about use [record_accessor plugin helper](https://docs.fluentd.org/plugin-helper-overview/api-plugin-helper-record_accessor) here?

lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -244,49 +248,102 @@ def reconnect_consumer
     end
   end
 
+  def process_batch_with_record_tag(batch)
+    es = {} 
+    batch.messages.each { |msg|
+      begin
+        record = @parser_proc.call(msg)
+        tag = record[@record_tag_key]
```
cosmo0920, line 256: Ditto.

cosmo0920 (commented):
I think that this patch shouldn't be supported nested record key on tag_source as `:record`.
Can we use record_accessor plugin helper to access record keys?

lib/fluent/plugin/in_kafka_group.rb
marratj: @cosmo0920 Not sure if we can use it. I basically copied the syntax from the existing `record_time_key` option. If we change this here, shouldn't then the syntax for `record_time_key` (lines 266-270) als be changed?

marratj: @cosmo0920 one more thing that just comes to my mind: when we are reading the records from Kafka at this stage, they are not yet Fluentd messages, so would the record_accessor helper even work?

cosmo0920: If we use Fluentd v1 API and handle nested records, it works. But this  plugin uses old Fluentd API. My suggestion does not work....

repeatedly:
Thanks!

Issue #287, opened 2019-08-22
function read_ssl_file doesn't support chained certificates


artur-fijalkowski:
This is not mentioned explicitly in README.md, but read_ssl_file function doesn't support chained certificates. In *NIX environment this is very common that *.pem file contains multiple certificates concatenated together (eg. root CA + intermediate CA).

https://github.com/fluent/fluent-plugin-kafka/blob/d0d9bbf51c9e5b16c18379917b3603a256262429/lib/fluent/plugin/kafka_plugin_util.rb#L35

This is (most probably) easy to fix by splitting file during read on BEGIN CERTIFICATE line.

repeatedly:
I'm not sure this sho uld be fixed in fluent-plugin-kafka or ruby-kafka.
Does this mean ruby-kafka supports chained certificates but fluent-plugin-kafka doesn't use it or ruby-kafka should accept chained certificates?

artur-fijalkowski:
I think that because ruby-kafka expects to get array of certificates and actual file read is in fluent-plugin-kafka the plugin is proper place for implementation. Also current documentation is a bit misleading. As it doesn't say anything about lack of chaining support (neither that ssl_ca_cert is an array of files).

syedriko:
There is a workaround: split the CA bundle file into individual PEM files and specify them as an array in fluent-plugin-kafka config:

ssl_ca_cert ["root_ca.crt", "intermediate_ca.crt"]

This takes advantage of https://github.com/fluent/fluent-plugin-kafka/blob/da73a50661c6e78f3e45f3e6aab535924aba9b09/lib/fluent/plugin/kafka_plugin_util.rb#L38

syedriko:
I'm working on a proper fix.

syedriko:
https://github.com/fluent/fluent-plugin-kafka/pull/390 is the PR tracking work on this issue, hopefully that'll un-stale it.

ashie:
Fixed by #410

PR #390, opened 2021-02-25
Issue 278: function read_ssl_file doesn't support chained certificates

syedriko:
Fix for https://github.com/fluent/fluent-plugin-kafka/issues/287, adding support for multiple PEM certificates in a single file.

### DIFF
--- a/lib/fluent/plugin/kafka_plugin_util.rb
+++ b/lib/fluent/plugin/kafka_plugin_util.rb
@@ -32,13 +32,27 @@ module Fluent
         end
       end
 
+      def invalid_path?(path)
+        path.nil? || path.respond_to?(:strip) && path.strip.empty?
+      end
+
       def read_ssl_file(path)
-        return nil if path.nil? || path.respond_to?(:strip) && path.strip.empty?
+        return nil if invalid_path?(path)
 
         if path.is_a?(Array)
           path.map { |fp| File.read(fp) }
         else
           File.read(path)
+        end
+      end
+
+      END_CERTIFICATE = "\n-----END CERTIFICATE-----\n"
+
+      def read_ssl_ca_certs
+        return nil if @ssl_ca_cert.nil?
+        valid_paths = @ssl_ca_cert.reject { |path| invalid_path?(path) }
+        valid_paths.flat_map do |path|
+          File.read(path).split(END_CERTIFICATE).map { |c| c += END_CERTIFICATE }
         end
       end
 
--- a/test/plugin/test_kafka_plugin_util.rb
+++ b/test/plugin/test_kafka_plugin_util.rb
@@ -8,7 +8,7 @@ class File
 end
 
 class KafkaPluginUtilTest < Test::Unit::TestCase
-    
+
     def self.config_param(name, type, options)
     end
     include Fluent::KafkaPluginUtil::SSLSettings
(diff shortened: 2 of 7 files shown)


syedriko:
@repeatedly Could you review this?

lib/fluent/plugin/kafka_plugin_util.rb
```diff
@@ -42,6 +46,13 @@ def read_ssl_file(path)
         end
       end
 
+      END_CERTIFICATE = "\n-----END CERTIFICATE-----\n"
+
+      def read_ssl_ca_certs
+        return nil if @ssl_ca_cert.nil?
+        @ssl_ca_cert.reject { |p| invalid_path?(p) }.flat_map { |c| File.read(c).split(END_CERTIFICATE).map { |c| c += END_CERTIFICATE} }
+      end
+
```
kenhys, line 55: It may be cosmetic feedback, but it is readable not to use oneliner aggressively.

```
valid_paths = @ssl_ca_cert.reject { |path| invalid_path?(path) }
valid_paths.flat_map do |path|
  File.read(path).split(END_CERTIFICATE).map { |c| c += END_CERTIFICATE}
end
```

syedriko: Corrected. Thanks!

jcantrill:
@kenhys ping... please review.  We could really use this fix in a formal release to resolve an open issue

ashie:
Note: Although [`Kafka::Client#initialize`](https://www.rubydoc.info/gems/ruby-kafka/Kafka/Client#initialize-instance_method) has an argument `ssl_ca_cert_file_path` to specify a path of a PEM encoded CA cert, it's not applicable in this case because `@ssl_ca_cert` can be an array of paths while the argument accept only one path.

lib/fluent/plugin/kafka_plugin_util.rb
```diff
@@ -42,6 +46,16 @@ def read_ssl_file(path)
         end
       end
 
+      END_CERTIFICATE = "\n-----END CERTIFICATE-----\n"
+
+      def read_ssl_ca_certs
+        return nil if @ssl_ca_cert.nil?
+        valid_paths = @ssl_ca_cert.reject { |path| invalid_path?(path) }
+        valid_paths.flat_map do |path|
+          File.read(path).split(END_CERTIFICATE).map { |c| c += END_CERTIFICATE }
```
ashie, line 55: Since it recognizes only `-----END CERTIFICATE-----`, it can't safely ignore other labels such as `RSA PRIVATE KEY`.
How about using `String#scan` like this:

```ruby
pem = File.read(path)
pem.scan(/-----BEGIN CERTIFICATE-----\n[\s\S]+?-----END CERTIFICATE-----\n/).to_a
```

test/plugin/test_kafka_plugin_util.rb
```diff
@@ -35,4 +35,17 @@ def test_read_ssl_file_when_non_empty_array
...
-end
\ No newline at end of file
+    def test_read_ssl_ca_certs_when_nil
+      assert_equal(nil, read_ssl_ca_certs())
+    end
+
+    def test_read_ssl_ca_certs_general_case
+      begin_cert = "-----BEGIN CERTIFICATE-----\n"
+      end_cert = "\n-----END CERTIFICATE-----\n"
+      cert_one = begin_cert + "one" + end_cert
```
ashie, line 45: I think only one line isn't suitable for a test.
Should use a more practical certificate string.

ashie (changes requested):

test/plugin/test_kafka_plugin_util.rb
ashie: A string actually generated by OpenSSL is desired.

jcantrill: how does anything other then the start and end certificate add value in a unit test?  The beginning and end lines are the only thing of significance since everything in between is irrelevent

ashie: At least intermediate data should be multi lines.
If we replace the implementation with `String#scan` and its regexp is inappropriate, it might not be able to match with multi lines even if it passes the test with a single line.

kenhys:
If it is ready, could you mark "resolve conversation" or re-request review again?

syedriko:
> If it is ready, could you mark "resolve conversation" or re-request review again?
I'm working on what I think is better fix, will submit as soon as.

syedriko:
https://github.com/fluent/fluent-plugin-kafka/pull/410 is a better fix for https://github.com/fluent/fluent-plugin-kafka/issues/287, but needs buy-in from ruby-kafka.

kenhys:
#410 approach seems better, I'll close this PR.

PR #398, opened 2021-04-14
Release v1.16.1

cosmo0920:
This version can be running with Ruby 3.

Our Ruby 3 use case should block Ruby 3 hash interop incompatibility.

Signed-off-by: Hiroshi Hatake <hatake@calyptia.com>


cosmo0920:
@ashie Could you kindly take a look?
Our Ruby 3 use case is blocked current released version which cannot run with Ruby 3.

ashie (approved):

ashie:
Thanks, I've released it: https://rubygems.org/gems/fluent-plugin-kafka/versions/0.16.1

cosmo0920:
Thanks a lot! :muscle:

PR #402, opened 2021-05-17
in_kafka: in_kafka_group: Support Ruby 3.0.0 keyword arguments interop

cosmo0920:
`@producer_opts` and `@consumer_opts` are also needed to hand-written
keyword argumtes interop on Ruby 3.

Signed-off-by: Hiroshi Hatake <hatake@calyptia.com>


ashie (approved):
Hmm, test for in_kafka doesn't exist...

cosmo0920:
Thanks.
We should add testcases for in_kafka....

Issue #397, opened 2021-04-12
Partition selection inconsistent with Java based Kafka tools 


ryanbamford:
We have fluentd setup to send events into kafka 
```
<match ** >
    @type kafka2

    brokers "#{ENV['KAFKA_BROKER_HOST']}:#{ENV['KAFKA_BROKER_PORT']}"
    sasl_over_ssl true
    username "#{ENV['KAFKA_BROKER_USERNAME']}"
    password "#{ENV['KAFKA_BROKER_PASSWORD']}"
    scram_mechanism "#{ENV['KAFKA_SCRAM_MECHANISM']}"
    ssl_ca_cert "#{ENV['KAFKA_CA_CERT_LOCATION']}"

    # topic rooting data
    default_topic        undefined-catchall  
 
    <format>
      @type json
    </format>

    <buffer topic>
      @type file
      path /fluentd/var/kafka-buffer/
      chunk_limit_size 2m
      queued_chunks_limit_size 500
      flush_mode interval
      flush_interval 10s
      flush_thread_count 20
      flush_at_shutdown true
      retry_max_interval 120
     
    </buffer>
    compression_codec gzip

    max_send_limit_bytes 2000000

  </match>
```
and are sending a message with 
```
"message_key":"mk-123456"
```
the same message is then being sent from the standard kafka-console-producer with the same message key. 
the message is being sent into the same kafka topic that has been set up with 25 partitions. 

the result we are seeing is 

* fluentd message -> Partition:3-mk-123456
* kConP message -> Partition:5-mk-123456

the messages are sent to different partitions ??? 

This is breaking the processing logic in kafka streams ... not all messages are showing up in the same consumers when partitions are being separated 

there seems to be a missmatch in the way the keys are being hashed. 

gem 'fluent-plugin-kafka', '0.16.0'
gem 'ruby-kafka', '1.3.0'
kafka -> strimzi kafka version 2.7.0

ryanbamford:
Think i can see the issue ... from the ruby kafka site 

Supported partitioning schemes
In order for semantic partitioning to work a partition_key must map to the same partition number every time. The general approach, and the one used by this library, is to hash the key and mod it by the number of partitions. There are many different algorithms that can be used to calculate a hash. By default crc32 is used. murmur2 is also supported for compatibility with Java based Kafka producers.

To use murmur2 hashing pass it as an argument to Partitioner. For example:
```
[...]

PR #405, opened 2021-05-26
Provide hash function choices

cosmo0920:
Fixes https://github.com/fluent/fluent-plugin-kafka/issues/397.

Kafka#new can handle Kafka::Partitioner classes which determine partitioner algorithms such as `crc32` and `murmur2`.
In Java based kafka client, default partitioner algorithm is murmur2 not crc32.
This default algorithm glitch causes different partion assignment between Java based Kafka client and Ruby based one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ If `ruby-kafka` doesn't fit your kafka environment, check `rdkafka2` plugin inst
       headers_from_record   (hash)   :default => {}
       use_default_for_unknown_topic (bool) :default => false
       discard_kafka_delivery_failed (bool) :default => false (No discard)
+      partitioner_hash_function (enum) (crc32|murmur2) :default => 'crc32'
 
       <format>
         @type (json|ltsv|msgpack|attr:<record name>|<formatter name>) :default => json
@@ -343,6 +344,7 @@ Support of fluentd v0.12 has ended. `kafka_buffered` will be an alias of `kafka2
       exclude_topic_key     (bool) :default => false
       exclude_partition_key (bool) :default => false
       get_kafka_client_log  (bool) :default => false
+      partitioner_hash_function (enum) (crc32|murmur2) :default => 'crc32'
 
       # See fluentd document for buffer related parameters: https://docs.fluentd.org/v/0.12/buffer
 
@@ -385,6 +387,7 @@ This plugin uses ruby-kafka producer for writing data. For performance and relia
       output_include_time   (bool) :default => false
       exclude_topic_key     (bool) :default => false
       exclude_partition_key (bool) :default => false
+      partitioner_hash_function (enum) (crc32|murmur2) :default => 'crc32'
 
       # ruby-kafka producer options
       max_send_retries    (integer) :default => 1
--- a/README.md
+++ b/README.md
@@ -231,6 +231,8 @@ If `ruby-kafka` doesn't fit your kafka environment, check `rdkafka2` plugin inst
 
 The `<formatter name>` in `<format>` uses fluentd's formatter plugins. See [formatter article](https://docs.fluentd.org/v/1.0/formatter).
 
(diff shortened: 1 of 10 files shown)


cosmo0920:
@ashie @kenhys Any comments? We should provide hash function choices because default hash algorithm differs from Java based Kafka client. This causes different partition assignment for similar event patterns. What do you think?

kenhys:
I'll take a look at it.

ashie:
Sorry, I can't have time to review it this week.

ashie:
I can review it next week.
(Do you need to hurry up?)

README.md
```diff
@@ -202,6 +202,7 @@ If `ruby-kafka` doesn't fit your kafka environment, check `rdkafka2` plugin inst
       headers_from_record   (hash)   :default => {}
       use_default_for_unknown_topic (bool) :default => false
       discard_kafka_delivery_failed (bool) :default => false (No discard)
+      partitioner_hash_function (enum) (crc32|murmur32) :default => 'crc32'
```
kenhys, line 206: Basically, giving a choice who can select hash algorithm is ok, but the default hash function may be 
better to change (from crc32 to murmur32 IMHO)

It breaks a compatibility, yes, so it should bump the version and add a note why the default partitioner hash
algorigthm is changed - it should follow the behavior of  kafka-console-producer for consistency.

repeatedly:
Adding note for `murmur32` is better. The explanation is like below:

Java based Kafka client uses `murmur32` by default. If you want to use same partitioning behavior with fluent-plugin-kafka, change it to `murmur32`.

lib/fluent/plugin/out_kafka_buffered.rb
```diff
@@ -133,15 +135,18 @@ def refresh_client(raise_error = true)
...
+                             sasl_plain_username: @username, sasl_plain_password: @password, sasl_over_ssl: @sasl_over_ssl, ssl_verify_hostname: @ssl_verify_hostname,
+                             partitioner: Kafka::Partitioner.new(hash_function: @partitioner_hash_function))
         else
           @kafka = Kafka.new(seed_brokers: @seed_brokers, client_id: @client_id, logger: logger, ssl_ca_cert: read_ssl_file(@ssl_ca_cert),
                              ssl_client_cert: read_ssl_file(@ssl_client_cert), ssl_client_cert_key: read_ssl_file(@ssl_client_cert_key), ssl_ca_certs_from_system: @ssl_ca_certs_from_system,
-                             sasl_gssapi_principal: @principal, sasl_gssapi_keytab: @keytab, sasl_over_ssl: @sasl_over_ssl,  ssl_verify_hostname: @ssl_verify_hostname)
+                             sasl_gssapi_principal: @principal, sasl_gssapi_keytab: @keytab, sasl_over_ssl: @sasl_over_ssl,  ssl_verify_hostname: @ssl_verify_hostname,
+                             partitioner: Kafka::Partitioner.new(hash_function: @partitioner_hash_function))
         end
```
kenhys, line 150: Could you add test cases to identify the behavior of the partitioner?

cosmo0920:
> I can review it next week.
> (Do you need to hurry up?)

I'm not hurry up to do this PR.

cosmo0920:
I happened to find that murmur2 hash algorithm on ruby-kafka is still unreleased for now....
https://github.com/zendesk/ruby-kafka/blob/65e4af54131405964d62a5d07d1c52d2a3018127/CHANGELOG.md#unreleased
I'm waiting for the new stable version of ruby-kafka.

cosmo0920:
I'd turned out this PR as Draft. if ruby-kafka releases partitioner keyword argument supported version, I'll turn on to be Ready for review.

cosmo0920:
I think that adding `digest-murmurhash` in fluent-agent bundled gem is enough for most users.

PR #407, opened 2021-06-10
Add option to create missing topic

fpytloun:


lib/fluent/plugin/out_kafka2.rb
```diff
@@ -286,6 +292,27 @@ def write(chunk)
           topic = @default_topic
           retry
         end
+        if @create_topic
+          producer.shutdown if producer
+          topic_template = chunk.metadata.variables[@topic_template_key_sym]
```
cosmo0920, line 297: `@topic_template_key.to_sym` ?

cosmo0920:
And DCO check is failed.
Under fluent organization, we use [DCO](https://github.com/fluent/fluent-plugin-kafka/pull/407/checks?check_run_id=2798234310).
Could you sign you commit(s)?

lib/fluent/plugin/out_kafka2.rb
cosmo0920: Oh, it's defined at https://github.com/fluent/fluent-plugin-kafka/pull/407/files#diff-d2728a692f5968218f3d1a0ab0eb4145cb76491cbc1a2a6ce697c84a4efcd1c7R175.

fpytloun:
Tested and seems to be working nicely :-)

cosmo0920 (commented):
Could you add an instruction for creating missing topic on README?

fpytloun:
Updated README.

fpytloun:
Hello, can this one get merged?

ashie:
I'll check it.

lib/fluent/plugin/out_kafka2.rb
```diff
@@ -151,6 +153,15 @@ def configure(conf)
         if @chunk_key_tag
           log.warn "default_topic is set. Fluentd's event tag is not used for topic"
         end
+        if @create_topic
+          raise Fluent::ConfigError, "Cannot set both use_default_for_unknown_topic and create_topic"
+        end
+      end
+
+      if @create_topic
+        if !@chunk_keys.include?(@topic_template_key)
```
ashie, line 162: Please use unless:
```diff
- if !@chunk_keys.include?(@topic_template_key)
+ unless @chunk_keys.include?(@topic_template_key)
```

README.md
```diff
@@ -286,6 +286,45 @@ If key name `partition_key_key` exists in a message, this plugin set the value o
...
+    topic_template ${tag.split('.')[0..1].join('.')}
+    types topic:string, topic_template:string
+  </record>
+</filter>
+
+<match app.**>
+  @type kafka2
+  create_topic          true
+  topic_template_key    topic_template
+  [...]
```
ashie, line 324: It seems that adding `<buffer>` section is more friendly to users because `topic_template_key` isn't added automatically:
```
<buffer topic_template,topic>
[...]
```

ashie (changes requested):

lib/fluent/plugin/out_kafka2.rb
```diff
@@ -151,6 +153,15 @@ def configure(conf)
         if @chunk_key_tag
           log.warn "default_topic is set. Fluentd's event tag is not used for topic"
         end
+        if @create_topic
+          raise Fluent::ConfigError, "Cannot set both use_default_for_unknown_topic and create_topic"
```
ashie, line 157: Probably it's not the code what you intended.
It should check `@use_default_for_unknown_topic` as described in the error message but the actual code checks `@default_topic.nil?`. Users cannot set `default_topic` due to this .

ashie: FYI: When I test this patch with the following the config, it causes crash due to empty topic (I intended to forward all unmatched events to `test` topic by `default_topic` but can't do it due to above issue):

```xml
<source>
  @type forward
  port 24224
  bind 0.0.0.0
</source>

<filter test.**>
  @type record_transformer
  auto_typecast true
  enable_ruby true
  <record>
    topic ${tag}
    topic_template ${tag.split('.')[0]}
    types topic:string, topic_template:string
  </record>
</filter>
 
<match **>
  @type copy
  <store>
    @type stdout
  </store>
  <store>
    @type kafka2
    create_topic true
    topic_template_key topic_template
    brokers localhost:9092
    #default_topic test # cannot set
    <format>
      @type json
    </format>
    <buffer topic_template,topic>
      flush_interval 10s
    </buffer>
  </store>
</match>
```

ashie (changes requested):

lib/fluent/plugin/out_kafka2.rb
fpytloun: It was meant not to be used together with default topic. Use should configure only one option - to use default topic for non-existing topics or to use topic template and create topics if it does not exist.

Otherwise code should be modified to use default topic if event does not have topic template in it's key. But in my opinion it's better to always ensure topic template or topic is set than relying on default topic option.

fpytloun: Pushed change that should allow it. Can you please check?

ashie: > Pushed change that should allow it. Can you please check?

Hmm, it doesn't the one I meant.
I meant that `default_topic` should be created in advance by a user, shouldn't created automatically by fluentd.

> But in my opinion it's better to always ensure topic template or topic is set than relying on default topic option.

I agree with you in this point.

On the other hand, forcedly rejecting `default_topic` is overkill, rejecting only `use_default_for_unknown_topic` is enough. We should give a user flexibility of the setting as possible, as I showed in above config (I don't intend to create `default_topic` automatically in this config, it's created in advance.) Since `default_topic` is `nil` by default, I think it doesn't conflict with the usage you meant.

BTW `default_topic` isn't only for `use_default_for_unknown_topic` but also used at the following, do you notice it?
https://github.com/fluent/fluent-plugin-kafka/blob/6fc25e112705a3e3f50c06d1a1157296ac4708a4/lib/fluent/plugin/out_kafka2.rb#L223
I intended to forward untagged events such as Fluentd's itself log to a pre-created topic.

ashie: Here is the patch for a1eef95861dd355257b331dec2c180eaf3d2e79e I meant
```diff
From b2cfc45947fa7413f2b10d5b7090e0fc1022d895 Mon Sep 17 00:00:00 2001
From: Takuro Ashie <ashie@clear-code.com>
Date: Fri, 20 Aug 2021 13:22:01 +0900
Subject: [PATCH] out_kafka2: Allow both create_topic & default_topic

In this case, default_topic should be created in advance, shouldn't
be created automatically.

Signed-off-by: Takuro Ashie <ashie@clear-code.com>
---
 lib/fluent/plugin/out_kafka2.rb | 7 ++++---
 1 file changed, 4 insertions(+), 3 deletions(-)

diff --git a/lib/fluent/plugin/out_kafka2.rb b/lib/fluent/plugin/out_kafka2.rb
index 0dad1ae..dbc80a5 100644
--- a/lib/fluent/plugin/out_kafka2.rb
+++ b/lib/fluent/plugin/out_kafka2.rb
@@ -153,9 +153,10 @@ DESC
         if @chunk_key_tag
           log.warn "default_topic is set. Fluentd's event tag is not used for topic"
         end
-        if @create_topic
-          raise Fluent::ConfigError, "Cannot set both use_default_for_unknown_topic and create_topic"
-        end
+      end
+
+      if @create_topic && @use_default_for_unknown_topic
+        raise Fluent::ConfigError, "Cannot set both use_default_for_unknown_topic and create_topic"
       end
 
       if @create_topic
-- 
2.30.2
```

(I tried to revert your second commit 328994a433ad598d0b8af81db4b0281fa12c3ba2 and add above, but I failed to push it since this pull request doesn't allowed to edit by us.)

fpytloun: I think my second commit does what you suggested - if topic does not exist and topic template key is not present in event (so topic cannot be created), it does fall back to writing into default topic.

kenhys:
It seems that whether which topic is used may be better to be resolved beforehand because of priority.
I mean in the following (mentioned already above)
(Determine it in  `rescue Kafka::UnknownTopicOrPartition` block is a bit late and it makes complicated.)

```
      topic = if @topic
                extract_placeholders(@topic, chunk)
              else
                (chunk.metadata.variables && chunk.metadata.variables[@topic_key_sym]) || @default_topic || tag
              end
```

PR #408, opened 2021-06-28
Update README.md

Konmoron:
Delete duplicate points


kenhys:
Thanks!

@Konmoron could you follow the instruction about DCO (signed off)
https://github.com/fluent/fluent-plugin-kafka/pull/408/checks?check_run_id=2929216721

PR #415, opened 2021-08-20
Enable excluding record fields

lucasMesquitaBorges:
I would like to exclude fields from message record.

This is desirable when using another configs such as `headers_from_record`, wich duplicates data to message headers. With this config will be possible to move data to headers and remove from the message payload.

PR #419, opened 2021-09-09
Exclude record fields

ashie:


ashie:
Sorry, I attempted to run CI on my forked repository to check the patch for #415 but I wrongly opened a new pull request at upstream...

PR #421, opened 2021-09-09
Add tests for rdkafka2

ashie:
Based on #415

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -5,6 +5,8 @@ on:
 jobs:
   build:
     runs-on: ${{ matrix.os }}
+    env:
+      USE_RDKAFKA: 1
     strategy:
       fail-fast: false
       matrix:
--- a/Gemfile
+++ b/Gemfile
@@ -2,3 +2,5 @@ source 'https://rubygems.org'
 
 # Specify your gem's dependencies in fluent-plugin-kafka.gemspec
 gemspec
+
+gem 'rdkafka', '>= 0.6.0' if ENV["USE_RDKAFKA"]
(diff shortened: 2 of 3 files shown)


fluent-plugin-kafka.gemspec
```diff
@@ -19,6 +19,7 @@ Gem::Specification.new do |gem|
   gem.add_dependency "fluentd", [">= 0.10.58", "< 2"]
   gem.add_dependency 'ltsv'
   gem.add_dependency 'ruby-kafka', '>= 1.4.0', '< 2'
+  gem.add_development_dependency "rdkafka", ">= 0.6.0"
```
cosmo0920, line 22: We shouldn't include rbkafka even if development dependency.
How about adding this in Gemfile instead?

cosmo0920 (changes requested):

fluent-plugin-kafka.gemspec
ashie: I've moved it to Gemfile and enable to switch by an environment variable `USE_RDKAFKA`.
BTW is there any obvious reason we it shouldn't be added to development dependency?

cosmo0920: > https://github.com/zendesk/ruby-kafka/pull/884#discussion_r568087727
> One of the ruby-kafka maintainer wants to be opt-in whenever C code included gem.

https://github.com/fluent/fluent-plugin-kafka/pull/405#discussion_r696322524

ashie: It's reasonable for digest-murmurhash because it's used with ruby-kafka.
But I don't think reasoble for rdkafka, because it's not used with ruby-kafka.

ashie: Anyway, rdkafka is optional for this plugin and I removed the dependency from gemspec.
Can I merge this?

cosmo0920 (approved):

fluent-plugin-kafka.gemspec
cosmo0920: Yes. Go ahead, please.

PR #426, opened 2021-09-27
out_rdkafka2: Add a new parameter `max_enqueue_bytes_per_second`

ashie:
When huge amount of messages are attempted to send, it causes excessive
load on both Fluentd and Kafka. Such situation might occure after down
time of Fluentd or Kafka.
Althogh it can be controled by combination of following parameters, it
isn't intuitive and might be hard to understand:

* Fluentd:
  * `flush_thread_burst_interval`
  * `buffer_queue_limit`
  * `buffer_chunk_limit`
* fluent-plugin-kafka
  * `enqueue_retry_backoff`
  * `max_enqueue_retries`
* librdkafka
  * `queue.buffering.max.kbytes`
  * `queue.buffering.max.messages`
  * `queue.buffering.max.ms`

This commit enable to control the enqueueing rate by a simple parameter.


ashie:
This pull request includes also a fix for a potential bug that `shutdown` method doesn't wait flush threads. I found the bug while I testing this feature.

test/plugin/test_out_rdkafka2.rb
```diff
@@ -116,5 +116,28 @@ def test_exclude_fields
       actual_messages = target_driver.events.collect { |event| event[2] }
       assert_equal([{'a' => 'b', 'message' => 'test'}], actual_messages)
     end
+
+    def test_max_enqueue_bytes_per_second
+      conf = config(default_topic: TOPIC_NAME) +
+             config_element('ROOT', '', {"max_enqueue_bytes_per_second" => 32 * 3,
+                                         "rdkafka_delivery_handle_poll_timeout" => 0}, [])
```
kenhys: how to conduct '32 * 3' as an appropriate bytes/second value?

ashie: I'm adjusting it in order to finish this test in 2-3 seconds.
In this test, all messages are 32bytes, so it's limited to send 3 messages per second.
Since this test send 9 messages in total

* 3 messages will be sent in 0-1 second
* 3 (total 6) messages will be sent in 1-2 seconds
* 3 (total 9) messages will be sent just after above

So that total test time is expected in 2-3 seconds.

ashie: (including staring & finishing the test driver)

kenhys: thank you for explanation, I got it.

kenhys (approved):
LGTM

cosmo0920 (changes requested):
Commits are really good. 
But, in fluent organization, we should add documentation if we add a new feature or parameter. 
It would be great if we add `max_enqueue_bytes_persecond` parameter on README.

And this feature might consume more CPU resource than not to be enabled.
We also should mention this on document.

ashie:
> we should add documentation if we add a new feature or parameter.

Oops, sorry I forgot to add it. I'll do it.

ashie:
> Commits are really good. But, in fluent organization, we should add documentation if we add a new feature or parameter. It would be great if we add `max_enqueue_bytes_persecond` parameter on README.

@cosmo0920 I've added it.

> And this feature might consume more CPU resource than not to be enabled. We also should mention this on document.

I think we don't have to worry about it.  Additional load by `EnqueueRate` is very low and `sleep` time is more longer than it in most cases.

ashie:
@cosmo0920 Can I merge this?

PR #437, opened 2021-11-25
Don't output skipped large events as warning log

ashie:
If `max_send_limit_bytes` is set, skipped large events are dumped as
warning log. It will fill up logs quickly and it might cause DoS.


fujimotos (approved):

fujimotos:
LGTM

PR #448, opened 2022-03-09
Add an option to support brokers' hostname with multiple addresses

akihiro17:
I'd like to use the option `resolve_seed_brokers` of `ruby-kafka` introduced in https://github.com/zendesk/ruby-kafka/pull/877 to support brokers' hostname with multiple addresses. This option is available since `ruby-kafka` 1.4.0.

ref. https://github.com/zendesk/ruby-kafka/releases/tag/v1.4.0

This PR adds an option `resolve_seed_brokers` and passes it to `Kafa#new`.


ashie:
Current required ruby-kafka is 1.4.0

ashie (approved):

akihiro17:
Thanks

PR #452, opened 2022-04-29
feat(out_rdkafka2): `sasl.username` and `sasl.password` is now configurable via plugin config

raytung:
While we were able to work around this by providing credentials via
`rdkafka_options` configuration parameter, this is not ideal since
`rdkafka_options` is not marked as a `secret` configuration parameter,
and thus the credentials will be logged to `stdout` during Fluentd start
up.

This commit makes it such that we are able to leverage the existing
`password` parameter that is already marked as secret.


lib/fluent/plugin/out_rdkafka2.rb
```diff
@@ -270,6 +270,8 @@ def build_config
       config[:"queue.buffering.max.messages"] = @rdkafka_buffering_max_messages if @rdkafka_buffering_max_messages
       config[:"message.max.bytes"] = @rdkafka_message_max_bytes if @rdkafka_message_max_bytes
       config[:"batch.num.messages"] = @rdkafka_message_max_num if @rdkafka_message_max_num
+      config[:"sasl.username"] = @username if @username
+      config[:"sasl.password"] = @password if @password
```
ashie, line 274: > This commit makes it such that we are able to leverage the existing
> password parameter that is already marked as secret.

Is it correct? out_rdkafka2 doesn't seem have these parameters.

ashie (changes requested):

lib/fluent/plugin/out_rdkafka2.rb
raytung: Ah yeah, if I'm understanding the code properly, we're including SASL related configuration on line 104 of out_rdkafka2, which should have username and password defined

ashie: Sorry, I missed it.

ashie (approved):

Issue #462, opened 2022-07-29
Newly introduced share_producer setting in 0.18.0 for out_kafka2 plugin results in BufferOverflowError


pchheda-lyft:
### Describe the bug

Producing to Kafka works as expected, with 0.17.5

However, with 0.18.0, producing to Kafka results in BufferOverflowError (essentially less performant?). This is true if share_producer setting is marked as true or even false

### To Reproduce

Flow of traffic ~7000 events/sec

`   
    @type kafka2
    # Hence, decreasing verbosity to warn
    # For more info, see https://docs.fluentd.org/deployment/logging
    brokers "#{ENV['KAFKA_BOOTSTRAP_SERVER']}"
    default_topic events
    # authentication settings
    ssl_ca_cert ["/code/lumosingest/kafka-certs/cert.2.pem", "/code/lumosingest/kafka-certs/cert.1.pem"]
    username "writer"
    password "#{ENV['PW']}"
    scram_mechanism "sha256"
    sasl_over_ssl true
    ssl_verify_hostname false
    # producer settings
    required_acks -1
    max_send_retries 10
    <inject>
      time_key collected_at
      time_type string
      time_format %Y-%m-%dT%H:%M:%S.%3NZ
    </inject>
    <format>
      @type json
    </format>
    get_kafka_client_log true

    max_send_limit_bytes 2850000

    <buffer events>
      flush_at_shutdown true
      flush_mode interval
      flush_interval 1s
      flush_thread_count 16
      queued_chunks_limit_size 16
      chunk_limit_size 3MB
      chunk_full_threshold 0.8
      total_limit_size 1024MB
      overflow_action block
    </buffer>
`

Have tried multiple settings, to increase flush_thread_count and queued_chunks_limit_size, chunk size and total limit size etc, to remove overflow_action as well.

this fails even in a staging environment with very less traffic.

Using this in conjunction with s3-sqs input plugin.

### Expected behavior

expected share_producer setting to work, without impact.

### Your Environment

```markdown
- Fluentd version: 1.15.1
- TD Agent version:
- fluent-plugin-kafka version: 0.18.0
- ruby-kafka version: 1.5.0
- Operating system: 
- Kernel version:
```

### Your Configuration

    @type kafka2
    # Hence, decreasing verbosity to warn
    # For more info, see https://docs.fluentd.org/deployment/logging
    brokers "#{ENV['KAFKA_BOOTSTRAP_SERVER']}"
    default_topic events
[...]

PR #466, opened 2022-08-14
fix(out_rdkafka2): fixed a bug where it is not respecting the Buffer sections' `chunk_limit_records` and `chunk_limit_size` options

raytung:
Fixes https://github.com/fluent/fluent-plugin-kafka/issues/462

Additionally, I just realised there's also a slight behavior change made in https://github.com/fluent/fluent-plugin-kafka/pull/454.

Prior to that commit, we were fetching Kafka topic related metadata only for the given topics, but that commit changes it such that the plugin fetches metadata for all topics. I don't think this will cause too much of an impact in production, but worth calling it out anyway.

### DIFF
--- a/lib/fluent/plugin/kafka_producer_ext.rb
+++ b/lib/fluent/plugin/kafka_producer_ext.rb
@@ -38,9 +38,15 @@ module Kafka
 end
 
 # for out_kafka2
+# Majority (if not all) of this code is lifted from https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/producer.rb
+# with the main difference where we have removed any checks regarding max_buffer_bytesize and max_buffer_size
+# The reason for doing this is to provide a better UX for our users where they only need to set those bounds in
+# the Buffer section using `chunk_limit_size` and `chunk_limit_records`.
+#
+# We should reconsider this in the future in case the `ruby-kafka` library drastically changes its internal.
 module Kafka
   class Client
-    def topic_producer(topic, compression_codec: nil, compression_threshold: 1, ack_timeout: 5, required_acks: :all, max_retries: 2, retry_backoff: 1, max_buffer_size: 1000, max_buffer_bytesize: 10_000_000, idempotent: false, transactional: false, transactional_id: nil, transactional_timeout: 60)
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/kafka_producer_ext.rb
```diff
@@ -38,9 +38,15 @@ def produce_for_buffered(value, key: nil, topic:, partition: nil, partition_key:
 end
 
 # for out_kafka2
+# Majority (if not all) of this code is lifted from https://github.com/zendesk/ruby-kafka/blob/master/lib/kafka/producer.rb
+# with the main difference where we have removed any checks regarding max_buffer_bytesize and max_buffer_size
+# The reason for doing this is to provide a better UX for our users where they only need to set those bounds in
+# the Buffer section using `chunk_limit_size` and `chunk_limit_records`.
+#
+# We should reconsider this in the future in case the `ruby-kafka` library drastically changes its internal.
```
raytung, line 46: This is what I was able to infer from the intent of this file, thought I should document it here so future us will know what this is doing. Let me know if this is incorrect.

cosmo0920 (approved):
Seems good. Great work! :+1:

ashie (approved):
LGTM, thanks for your fix!

@raytung BTW are you interested in maintaining this plugin?
If so, I want to invite you to this project.

lib/fluent/plugin/kafka_producer_ext.rb
ashie: It seems correct.
@repeatedly Please comment on this if you have any concerns.

raytung:
Hey @ashie for sure, I'm happy to help out whenever I can, but I can't promise to be super on top of everything due to life / work etc if that's okay.

ashie:
I've Invited you. Please check e-mail.

> Hey @ashie for sure interested, I'm happy to help out whenever I can! However, I can't promise to be super on top of everything due to life / work etc if that's okay.

Yeah of course it's O.K., we Fluentd maintenance team also continue to maintain it.
Since our man-power isn't always enough to maintain all plugins, your contribution is very helpful for us!
Thanks a lot for joining us!

lib/fluent/plugin/kafka_producer_ext.rb
repeatedly: This comment is correct. I implemented this class to remove some overhead/unmatched parts from ruby-kafka's producer.

ashie: Thanks for your comment!

Issue #463, opened 2022-08-01
rdkafka 0.12.0 is incompatible with fluent-plugin-kafka v0.18.0 or former


bertpassek:
### Describe the bug

We noticed a null pointer exception when td-agent is being restartet. But so far everything is working properly, no side effects , just this exxception stracktrace.

### To Reproduce

restart td-agent

### Expected behavior

no exception

### Your Environment

```markdown
- Fluentd version: 1.15.1
- TD Agent version:
- Operating system: 20.04.4 LTS (Focal Fossa)
- Kernel version: 5.13.0-1031-aws
```

### Your Configuration

```apache
<match home.tracking.**>
    @type "rdkafka2"
    @id "out_kafka"
    @log_level "warn"
    brokers "..."
    use_event_time "true"
    topic_key "topic"
    default_topic "unknown"
    required_acks 1
    max_enqueue_retries 60
    enqueue_retry_backoff 5
    compression_codec "snappy"
    rdkafka_buffering_max_ms 10
    rdkafka_buffering_max_messages 10000000
    rdkafka_message_max_num 1000
    max_send_retries 2147483647
    ack_timeout 60
    rdkafka_delivery_handle_poll_timeout 0

    <format >
      @type "json"
    </format>


    <buffer topic>
      @type "file"
      chunk_limit_size "256MB"
      flush_interval "5s"
      flush_at_shutdown "true"
      flush_thread_count 4
      retry_forever "true"
      retry_wait "5s"
      total_limit_size "2G"
      overflow_action "throw_exception"
      compress "gzip"
    </buffer>

  </match>
```

### Your Error Log

```shell
2022-08-01 10:56:20 +0000 [warn]: #0 unexpected error while shutting down on output plugin plugin=Fluent::Rdkafka2Output plugin_id="out_kafka" error_class=NoMethodError error="undefined method `join' for nil:NilClass"
  2022-08-01 10:56:20 +0000 [error]: #0 /var/lib/gems/2.7.0/gems/fluent-plugin-kafka-0.18.0/lib/fluent/plugin/out_rdkafka2.rb:330:in `block (3 levels) in shutdown_producers'
  2022-08-01 10:56:20 +0000 [warn]: #0 /var/lib/gems/2.7.0/gems/fluent-plugin-kafka-0.18.0/lib/fluent/plugin/out_rdkafka2.rb:15:in `close'
[...]

PR #468, opened 2022-08-17
fix(out_rdkafka2): fix compatibility with rdkafka 0.12.0

raytung:
Fixes https://github.com/fluent/fluent-plugin-kafka/issues/463

I think this is the simplest implementation that I could find to support rdkafka 0.12.0, but unfortunately because of how we're monkey patching the methods, it means we'd have to drop support for < 0.12.0 thus making this a breaking change.

Not sure if this is the right place to propose whether we should consider splitting out rdkafka into its own plugin (say `fluent-plugin-rdkafka`) so we're able to put the rdkafka gem as a dependency in the gemspec and give it a proper version range.

The "test" for this change is that I have created a separate PR that only bumps rdkafka gem to 0.12.0 and nothing else. https://github.com/fluent/fluent-plugin-kafka/pull/469, then check its unit test logs https://github.com/fluent/fluent-plugin-kafka/runs/7878417388?check_suite_focus=true#step:5:1030

Comparing to the test logs for this PR, https://github.com/fluent/fluent-plugin-kafka/runs/7878310221?check_suite_focus=true, we don't see any more `NoMethodError:undefined method 'join' for nil:NilClass` errors.

### DIFF
--- a/lib/fluent/plugin/out_rdkafka2.rb
+++ b/lib/fluent/plugin/out_rdkafka2.rb
@@ -30,7 +30,7 @@ class Rdkafka::Producer
   def close(timeout = nil)
     rdkafka_version = Rdkafka::VERSION || '0.0.0'
     # Rdkafka version >= 0.12.0 changed its internals
-    if rdkafka_version.split('.')[1].to_i >= 12
+    if Gem::Version::create(rdkafka_version) >= Gem::Version.create('0.12.0')
       ObjectSpace.undefine_finalizer(self)
 
       return @client.close(timeout)
--- a/lib/fluent/plugin/out_rdkafka2.rb
+++ b/lib/fluent/plugin/out_rdkafka2.rb
@@ -5,9 +5,37 @@ require 'fluent/plugin/kafka_plugin_util'
 
 require 'rdkafka'
 
+# This is required for `rdkafka` version >= 0.12.0
+# Overriding the close method in order to provide a time limit for when it should be forcibly closed
+class Rdkafka::Producer::Client
+  # return false if producer is forcefully closed, otherwise return true
+  def close(timeout=nil)
+    return unless @native
+
+    # Indicate to polling thread that we're closing
+    @polling_thread[:closing] = true
+    # Wait for the polling thread to finish up
+    thread = @polling_thread.join(timeout)
+
+    Rdkafka::Bindings.rd_kafka_destroy(@native)
+
+    @native = nil
+
+    return !thread.nil?
+  end
+end
+
 class Rdkafka::Producer
   # return false if producer is forcefully closed, otherwise return true
   def close(timeout = nil)
+    rdkafka_version = Rdkafka::VERSION || '0.0.0'
+    # Rdkafka version >= 0.12.0 changed its internals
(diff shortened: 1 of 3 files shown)


lib/fluent/plugin/out_rdkafka2.rb
```diff
@@ -5,19 +5,31 @@
...
   # return false if producer is forcefully closed, otherwise return true
   def close(timeout = nil)
-    @closing = true
-    # Wait for the polling thread to finish up
-    # If the broker isn't alive, the thread doesn't exit
-    if timeout
-      thr = @polling_thread.join(timeout)
-      return !!thr
-    else
-      @polling_thread.join
```
raytung, line 18: rdkafka gem changed its internals from this commit on wards https://github.com/appsignal/rdkafka-ruby/commit/a391d26f0729fa395974165a03e9aef3e8c4486b

ashie:
> Not sure if this is the right place to propose whether we should consider splitting out rdkafka into its own plugin (say fluent-plugin-rdkafka) so we're able to put the rdkafka gem as a dependency in the gemspec and give it a proper version range.

I'm also considering this option.
Currently rdkafka isn't included in dependent gem so we can't even block rdkafka 0.12.0 or later.
We should consider how should be realize it since some codes are shared between out_kafka2 and out_rdkafka.
Instead of creating a new repository, adding a new gemspec file might be enough for this.

> I think this is the simplest implementation that I could find to support rdkafka 0.12.0, but unfortunately because of how we're monkey patching the methods, it means we'd have to drop support for < 0.12.0 thus making this a breaking change.

Another option is that including both code and switching them by checking rdkafka's version.
Probably it can be checked by `Rdkafka::VERSION`:
https://github.com/appsignal/rdkafka-ruby/blob/c02f217a189b71e3e253fe4ad59286b5fc9d4034/lib/rdkafka/version.rb#L4

raytung:
> Instead of creating a new repository, adding a new gemspec file might be enough for this.

Ah yeah, that will probably be the simpler option!

> Another option is that including both code and switching them by checking rdkafka's version.

This could work. Let me play around with this idea.

raytung:
Hey @ashie, as far as I can tell, the only reason for monkey patching the `close` method is so that we can have a hard timeout when closing rdkafka's producer client.

https://github.com/fluent/fluent-plugin-kafka/pull/468/files#diff-357aa79f6db32ff5e634032b9c03c05fa1c626bc165938f3a0f74385e45fe8ccR369-R373

Would it be problematic if we just wait for the thread to shutdown by itself? If there's no foreseeable problems with that, this patch could be a lot simpler.

ashie:
I'm sorry for not responding to this.
I haven't forgotten this issue, but I'm quite busy and hard to check the details yet.

ashie:
Sorry for my late response.
Now I'm going to release v0.19 with this patch (with some nitpick fixes).

> Would it be problematic if we just wait for the thread to shutdown by itself? If there's no foreseeable problems with that, this patch could be a lot simpler.

I'm not sure, so that the current patch seems reasonable for safety.

ashie (approved):
LGTM

ashie:
Thanks for your effort!

PR #474, opened 2022-12-19
GitHub issue: Set waiting-for-triage by default

ashie:
We might not check issues so often recent days, so it is inappropriate to close stale issues that aren't checked by any maintainers. This commit will add waiting-for-triage label to all new issues and won't close them by default. In addition, issues that are assigned or set milestone are also exempted.


kenhys (approved):
LGTM.

Issue #472, opened 2022-10-25
Support AWS MSK Authentication


nonamef:
### Is your feature request related to a problem? Please describe.

Currently using fluent with AWS MSK as unauthenticated with security groups to control access. Want to enable AWS IAM assume role for cross account access.

### Describe the solution you'd like

Same setup as fluent-plugin-s3 for **assume_role_credentials** would be good.
e.g https://github.com/fluent/fluent-plugin-s3/blob/master/docs/credentials.md

### Describe alternatives you've considered

Only alternative is to continue as unauthenticated.

raytung:
Hey @nonamef We have added AWS IAM auth support with access keys and secret keys. Unfortunately, since the `ruby-kafka` gem is deprecated, they are not adding support for assume roles. Looks like librdkafka is also not supporting AWS IAM auth so our hands are a bit tied on this one.

PR #481, opened 2023-01-21
feat(out_kafka2): adds support for AWS IAM authentication to MSK usin…

raytung:
…g long lived credentials

Addresses https://github.com/fluent/fluent-plugin-kafka/issues/472

This PR adds support for AWS IAM authentication using long lived credentials (access key id and secret access keys). To support AWS assume role and STS, we will need to wait for upstream's `ruby-kafka` library support.

We will need to bump `ruby-kafka` to 1.5.0 in order to support this feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,6 +212,16 @@ If `ruby-kafka` doesn't fit your kafka environment, check `rdkafka2` plugin inst
       partitioner_hash_function (enum) (crc32|murmur2) :default => 'crc32'
       share_producer        (bool)   :default => false
 
+      # If you intend to rely on AWS IAM auth to MSK with long lived credentials
+      # https://docs.aws.amazon.com/msk/latest/developerguide/iam-access-control.html
+      #
+      # For AWS STS support, see status in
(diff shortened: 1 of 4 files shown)


ashie (approved):
LGTM
Thanks!

ashie:
It's not this PR's problem but I think it would be better to unify setting mandatory parameters.

ashie:
> we will need to wait for upstream's `ruby-kafka` library support.

BTW according to https://github.com/zendesk/ruby-kafka ruby-kafka is no longer maintained.
We might need to consider deprecating also this plugin...

raytung:
@ashie I saw that too 😢 I spent a bit of time digging up the story for librdkafka but it looks like there's a heavy push back from the author / Confluent to support pluggable authentication call back in librdkafka which means there's no way to support AWS IAM auth 😢 (Basically they are not implementing this KIP https://cwiki.apache.org/confluence/display/KAFKA/KIP-86%3A+Configurable+SASL+callback+handlers)

Issue #489, opened 2023-05-15
Implement `use_default_for_unknown_topic` for rdkafka2 output plugin


raytung:
### Is your feature request related to a problem? Please describe.

Currently the rdkafka2 output plugin has a `default_topic` configuration but that's only for when the message itself does not contain a topic field. If the `default_topic` parameter is not set, and that the plugin tries to write to a topic that does not exist in Kafka brokers, the plugin will produce a log `Local: Unknown topic (unknown_topic)` and then stops processing.

### Describe the solution you'd like

Implement `use_default_for_unknown_topic` for rdkafka2 output plugin, similar to the parameter for the kafka2 output plugin, where it will write to the default topic if the desired topic is not found in Kafka brokers.

If the default topic does not exist in Kafka brokers, we should then log a warning message. (Possibly allow configuring whether to throw exception or block or drop the message).

A naive implementation could be identical to the solution in kafka2 output:
1. We first attempt to write to the desired topic
2. Rescue the producer produce attempt, and if the rescued error is unknown topic, we set override the topic name of the message and retry. (This is done in this code  https://github.com/fluent/fluent-plugin-kafka/blob/c21236f20e23226c166a93e38d2c24401fa59d90/lib/fluent/plugin/out_rdkafka.rb#L309-L320, and check whether the error code equals to `:unknown_topic`

A potential optimization could be:
First check whether the topic exist in the topic metadata before attempting produce. I'm not familiar enough with librdkafka to know whether it does this already.

### Describe alternatives you've considered

N/A

ashie:
We'll keep this opened untill https://github.com/fluent/fluent-plugin-kafka/pull/490#discussion_r1193900681 is resolved.

raytung:
Documenting for visibility, #490 implemented the logic to publish to the default topic if topic is not found in Kafka brokers. However, we have not considered publishing to the default topic if we receive a `UNKNOWN_PARTITION` error (the `out_kafka2` plugin does this).

I can think of a couple of cases where we would receive an `UNKNOWN_PARTITION` error (potentially more):
1. Kafka broker had corrupted metadata (or topic data has not propagated to all brokers yet)
2. The topic was removed in Kafka and then a new topic with the same name is re-created but with less number of partitions
3. Producer does not have proper read access

In all cases, I think it makes sense to publish to the default topic, but perhaps we should `warn` log. What do you think @ashie?

ashie:
Thanks for clarification for this.

[...]

PR #490, opened 2023-05-15
feat(out_rdkafka2): adds `use_default_for_unknown_topic` configuration

raytung:
Addresses https://github.com/fluent/fluent-plugin-kafka/issues/489

This lifts the same parameter from the `out_kafka2` plugin and behaves similarly - it first attempts to write to a given topic, and if it fails because the topic does not exist, it will attempt to write to the `default_topic`

### DIFF
--- a/lib/fluent/plugin/out_rdkafka2.rb
+++ b/lib/fluent/plugin/out_rdkafka2.rb
@@ -65,6 +65,7 @@ DESC
     config_param :topic_key, :string, :default => 'topic', :desc => "Field for kafka topic"
     config_param :default_topic, :string, :default => nil,
                  :desc => "Default output topic when record doesn't have topic field"
+    config_param :use_default_for_unknown_topic, :bool, :default => false, :desc => "If true, default_topic is used when topic not found"
     config_param :message_key_key, :string, :default => 'message_key', :desc => "Field for kafka message key"
     config_param :default_message_key, :string, :default => nil
     config_param :partition_key, :string, :default => 'partition', :desc => "Field for kafka partition"
@@ -233,6 +234,9 @@ DESC
       @rdkafka = Rdkafka::Config.new(config)
 
       if @default_topic.nil?
+        if @use_default_for_unknown_topic
+          raise Fluent::ConfigError, "default_topic must be set when use_default_for_unknown_topic is true"
+        end
         if @chunk_keys.include?(@topic_key) && !@chunk_key_tag
           log.warn "Use '#{@topic_key}' field of event record for topic but no fallback. Recommend to set default_topic or set 'tag' in buffer chunk keys like <buffer #{@topic_key},tag>"
         end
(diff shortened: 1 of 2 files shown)


lib/fluent/plugin/out_rdkafka2.rb
```diff
@@ -466,24 +470,41 @@ def write(chunk)
...
               sleep @enqueue_retry_backoff
               attempt += 1
             else
               raise "Failed to enqueue message although tried retry #{@max_enqueue_retries} times"
             end
+          # https://github.com/confluentinc/librdkafka/blob/c282ba2423b2694052393c8edb0399a5ef471b3f/src/rdkafka.h#LL309C9-L309C41
+          # RD_KAFKA_RESP_ERR__UNKNOWN_TOPIC
+          when :unknown_topic
+            if @use_default_for_unknown_topic && actual_topic != @default_topic
+              log.debug "'#{actual_topic}' topic not found. Retry with '#{@default_topic}' topic"
```
raytung, line 503: Note: this is one major difference to the `out_kafka2` plugin's implementation - we are debugging logging instead of warn logging. This is because it is an intended behavior if we do get into this code path (since `use_default_for_unknown_topic` is set!). We should consider publishing librdkafka's own internal metrics though, but this should be discussed in a separate issue if we're not already doing this.

lib/fluent/plugin/out_rdkafka2.rb
```diff
@@ -466,24 +470,41 @@ def write(chunk)
...
             if attempt <= @max_enqueue_retries
               log.warn "Failed to enqueue message; attempting retry #{attempt} of #{@max_enqueue_retries} after #{@enqueue_retry_backoff}s"
               sleep @enqueue_retry_backoff
               attempt += 1
             else
               raise "Failed to enqueue message although tried retry #{@max_enqueue_retries} times"
             end
+          # https://github.com/confluentinc/librdkafka/blob/c282ba2423b2694052393c8edb0399a5ef471b3f/src/rdkafka.h#LL309C9-L309C41
+          # RD_KAFKA_RESP_ERR__UNKNOWN_TOPIC
+          when :unknown_topic
```
raytung, line 501: Note: another major difference to the `out_kafka2`'s implementation - the `out_kafka2` plugin also writes to the `default_topic` if we receive an `UNKNOWN_PARTITION` error, which I have left out in this implementation since I'd like to consider a bit more on whether we should publish to `default_topic` or whether we should have a separate configuration to handle that error.

ashie: Note: https://github.com/fluent/fluent-plugin-kafka/blob/c21236f20e23226c166a93e38d2c24401fa59d90/lib/fluent/plugin/out_kafka2.rb#L403-L408

lib/fluent/plugin/out_rdkafka2.rb
ashie: >  This is because it is an intended behavior if we do get into this code path (since use_default_for_unknown_topic is set!). 

I agree with you, the log level should be debug.

ashie (approved):

ashie:
Thanks!

PR #491, opened 2023-05-27
feat(out_rdkafka2): add parameter `use_default_for_unknown_partition_error`

raytung:
Implements suggestion in https://github.com/fluent/fluent-plugin-kafka/issues/489#issuecomment-1558405103


lib/fluent/plugin/out_rdkafka2.rb
```diff
@@ -504,6 +505,16 @@ def enqueue_with_retry(producer, topic, record_buf, message_key, partition, head
               actual_topic = @default_topic
               retry
             end
+            raise e
+          # https://github.com/confluentinc/librdkafka/blob/c282ba2423b2694052393c8edb0399a5ef471b3f/src/rdkafka.h#L305
+          # RD_KAFKA_RESP_ERR__UNKNOWN_PARTITION
+          when :unknown_partition
+            if @use_default_for_unknown_partition_error && actual_topic != @default_topic
+              log.debug "failed writing to topic '#{actual_topic}' with error '#{e.to_s}'. Writing message to topic '#{@default_topic}'"
```
ashie, line 513: Shouldn't it `log.warn` for this according to your comment in https://github.com/fluent/fluent-plugin-kafka/issues/489#issuecomment-1558405103?

raytung: yeah I thought about it some more - because we require a config parameter to enter into this logic, I think it makes sense to debug log this instead since it is doing what it intends to do. And in the event where this config parameter is not set (default to false), it will simply bubble the exception upwards which will be logged by fluentd anyway.

ashie: Yep, it's reasonable.
Thanks for clarification.

ashie:
Thanks!

PR #496, opened 2023-10-10
Enhancement : Add "discard_kafka_delivery_failed_regex" option

kubotat:
When with Forwarder and Aggregator architecture, forwarders sometimes send invalid data which cause delivery failure at aggregators. `out_rdkafka2` plugin has`discard_kafka_delivery_failed` but it potentially discards not only unnecessary events but also required events, e.g. users expect Fluentd to keep events in buffer files during the outage of Kafka cluster (due to maintenance for instance) but all events are discarded when with `discard_kafka_delivery_failed`.
This enhancement request proposes having `discard_kafka_delivery_failed_regex` option on `out_rdkafka2` plugin to nullify invalid data by checking the error message with given regexp pattern.

Here is a sample use case:
In the following configuration, dummy events are generated with tag `test-topic0001` and Fluentd try to ship message to `test-topic0001`. If `test-topic0001` does not exist in Kafka cluster, `out_rdkafka2` emits a `Local: Unknown topic (unknown_topic)` warning message. With `discard_kafka_delivery_failed_regex /Unknown topic/`, `out_rdkafka2` discards events which emits a `Local: Unknown topic (unknown_topic)` warning message.

```
<source>
  @type dummy
  sample {"Hello":"World"}
  tag test-topic0001
</source>

<match dummy>
    @type rdkafka2
    @id "rdkafka2"
    bokers kafka01.demo.local:9093, kafka02.demo.local:9093
    discard_kafka_delivery_failed_regex /Unknown topic/
    required_acks 1
    ack_timeout 20
    share_producer true
    ## Buffer settings
    <buffer tag>
      @type file
      path ./buffer/
      flush_interval 5s
    </buffer>
    <format>
      @type json
    </format>
</match>
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -517,6 +517,7 @@ You need to install rdkafka gem.
       exclude_topic_key     (bool) :default => false
       exclude_partition_key (bool) :default => false
       discard_kafka_delivery_failed (bool) :default => false (No discard)
+      discard_kafka_delivery_failed_regex (regexp) :default => nil (No discard)
       use_event_time        (bool) :default => false
 
       # same with kafka2
@@ -558,6 +559,9 @@ You need to install rdkafka gem.
       # to send. Default is no limit.
       max_enqueue_bytes_per_second (integer) :default => nil
     </match>
+
+`rdkafka2` supports `discard_kafka_delivery_failed_regex` parameter:
+- `discard_kafka_delivery_failed_regex` - default: nil - discard the record where the Kafka::DeliveryFailed occurred and the emitted message matches the given regex pattern, such as `/unknown_topic/`. 
 
 If you use v0.12, use `rdkafka` instead.
 
--- a/lib/fluent/plugin/out_rdkafka2.rb
+++ b/lib/fluent/plugin/out_rdkafka2.rb
@@ -65,6 +65,8 @@ DESC
     config_param :topic_key, :string, :default => 'topic', :desc => "Field for kafka topic"
     config_param :default_topic, :string, :default => nil,
                  :desc => "Default output topic when record doesn't have topic field"
+    config_param :use_default_for_unknown_topic, :bool, :default => false, :desc => "If true, default_topic is used when topic not found"
+    config_param :use_default_for_unknown_partition_error, :bool, :default => false, :desc => "If true, default_topic is used when received unknown_partition error"
     config_param :message_key_key, :string, :default => 'message_key', :desc => "Field for kafka message key"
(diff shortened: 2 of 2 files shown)


ashie:
It seems nice, thanks for your proposal!
To merge this, could you address following things?

* Could you follow DCO?: https://github.com/fluent/fluent-plugin-kafka/pull/496/checks?check_run_id=17580570217
* Could you add a description about this parameter to README.md?
* (It would be nice if we can add a test for it, but not mandatory for now.)

As a future task for us, it would be nice if we can match errors by error codes of Rdkafka to avoid ambiguity: https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafkacpp_8h.html#a4c6b7af48c215724c323c60ea4080dbf
but it's not needed in this pull request.

ashie (changes requested):
It seems nice, thanks for your proposal!
To merge this, could you address following things?

* Could you follow DCO?: https://github.com/fluent/fluent-plugin-kafka/pull/496/checks?check_run_id=17580570217
* Could you add a description about this parameter to README.md?
* (It would be nice if we can add a test for it, but not mandatory for now.)

As a future task for us, it would be nice if we can match errors by error codes of Rdkafka to avoid ambiguity: https://docs.confluent.io/platform/current/clients/librdkafka/html/rdkafkacpp_8h.html#a4c6b7af48c215724c323c60ea4080dbf
but it's not needed in this pull request.

kubotat:
@ashie Thanks for reviewing. I've done two items you mentioned. I will keep working on a test code.
- [x] Could you follow DCO?: https://github.com/fluent/fluent-plugin-kafka/pull/496/checks?check_run_id=17580570217
- [x] Could you add a description about this parameter to README.md?
- [] (It would be nice if we can add a test for it, but not mandatory for now.)

It seems some of the check processes were failed after updating the contents. I would appreciate it if you could give me an advice on how to pass the check process.

ashie:
> It seems some of the check processes were failed after updating the contents. I would appreciate it if you could give me an advice on how to pass the check process.

Don't worry, they aren't caused by this change.
They are always unstable, need to rerun when failed.
We should tackle on this as another issue.

> I will keep working on a test code.

Please let me know if it's hard to add. I'll merge & release it without test for now.

kubotat:
@ashie Thanks. I would appreciate it if could merge my request.

ashie (approved):

ashie:
Thanks!

ashie:
I've released v0.19.2

Issue #500, opened 2023-11-27
ruby-kafka is no longer maintained and superseded by librdkafka via rdkafka-ruby bindings


mensfeld:
### Describe the bug

I just wanted to raise to your attention that based on the official statement from the ruby-kafka repo:

> This library is no longer actively developed and has been superseded by [librdkafka](https://github.com/edenhill/librdkafka) via [rdkafka-ruby](https://github.com/appsignal/rdkafka-ruby/) bindings. While this library may still receive security patches and bug fixes, it is no longer recommended for production usage.

ruby-kafka has some critical errors that can lead to data loss. Even zendesk already switched to it with their other operations.

While I cannot replace it fully (rdkafka and waterdrop maintainer here) in this plugin because of my limited knowledge, I am happy to help you guys out :)

The downside is jruby - which is something that will have to be figured out.

### To Reproduce

To reproduce or rather see, check the `.gemspec`

### Expected behavior

I would expect this library to use one of the maintained and recommended Kafka gem.

### Your Environment

```markdown
Applies to any environment.
```

### Your Configuration

Applies to any config.

### Your Error Log

```shell
Irrelevant.
```

### Additional context

Please close it if you feel this is not a bug and it should be raised as a discussion or something else. From my perspective it is a bug as it poses risk of data loss for the end users.

ashie:
Yes, that's been a concern for us recently.
We already have a rdkafka based plugin: https://github.com/fluent/fluent-plugin-kafka/blob/master/lib/fluent/plugin/out_rdkafka2.rb so one of choice is dropping out_kafka2 plugin and maintain only out_rdkafka2.

About waterdrop I've just found out it now.
It might be nice to replace ruby-kafka. Thanks for notifying it!
We'll consider about it.

mensfeld:
Your implementation will cause Ruby VM crashes for sure in one place: you cannot destroy rdkafka object on timeout during close if there are ANY outgoing messages. Both this and the events queue needs to be purged: https://github.com/karafka/waterdrop/blob/master/lib/waterdrop/producer.rb#L186

otherwise once in a while there will be dangling data that librdkafka does not tolerate and causes VM failures.

Anyhow keep me posted and happy to help :)

raytung:
[...]

PR #505, opened 2023-12-09
out_rdkafka2: add patch for new version of rdkafka

ttych:
All PR are failing now because of rdkafka gem update.

The current code works with rdkafka 0.12.0,
and in 0.14.0 (there is no 0.13.0), the `Rdkafka::Producer` implementation has changed,
no more `@client` and `Rdkafka::Producer::Client`, 
now there is a `@native_client` and `Rdkafka::NativeKafka`.

No update on this part of the code between 0.14.0 and 0.15.0.

There is a constraint in the out_rdkafka2, on the `Rdkafka::Producer#close`, to add a timeout,
which is not part of the standard gem.

I try to update this behavior since following the current way to patch.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -509,7 +509,6 @@ You need to install rdkafka gem.
       partition_key         (string) :default => 'partition'
       partition_key_key     (string) :default => 'partition_key'
       message_key_key       (string) :default => 'message_key'
-      default_topic         (string) :default => nil
       use_default_for_unknown_topic           (bool) :default => false
       use_default_for_unknown_partition_error (bool) :default => false
       default_partition_key (string) :default => nil
--- a/lib/fluent/plugin/rdkafka_patch/0_14_0.rb
+++ b/lib/fluent/plugin/rdkafka_patch/0_14_0.rb
@@ -1,0 +1,44 @@
+class Rdkafka::NativeKafka
+  # return false if producer is forcefully closed, otherwise return true
+  def close(timeout=nil, object_id=nil)
+    return true if closed?
+
+    synchronize do
+      # Indicate to the outside world that we are closing
+      @closing = true
+
+      thread_status = :unknown
+      if @polling_thread
+        # Indicate to polling thread that we're closing
+        @polling_thread[:closing] = true
+
+        # Wait for the polling thread to finish up,
+        # this can be aborted in practice if this
+        # code runs from a finalizer.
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -509,7 +509,6 @@ You need to install rdkafka gem.
       partition_key         (string) :default => 'partition'
       partition_key_key     (string) :default => 'partition_key'
       message_key_key       (string) :default => 'message_key'
-      default_topic         (string) :default => nil
```
ttych, line 512: This line in conf block appears twice,
I just removed one.

ttych:
I just see #500,
so this is not a safe way to do it this way.

and also, waterdrop encapsulation for rdkafka is just for producer side.

lib/fluent/plugin/rdkafka_patch/0_14_0.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+end
+
+class Rdkafka::Producer
+  def close(timeout = nil)
+    return true if closed?
+    ObjectSpace.undefine_finalizer(self)
+    @native_kafka.close(timeout)
+  end
+end
```
ashie, line 44: This part is needed to update to follow rdkafka-ruby v0.16: 
https://github.com/karafka/rdkafka-ruby/blob/cff293df9debcb4843353dbd92bf08c146fb8188/lib/rdkafka/producer.rb#L139-L151

ashie (commented):
Sorry for my late response, I overlooked this.
It looks good, but need to update for rdkafka-ruby v0.16.

Hey @kenhys, I'm now considering to revert #513 then apply this with some further improvements. In addition, we may need to feedback this timeout patch to upstream to avoid future monkey patching. What do you think about it?

kenhys:
I'm not sure yet but I agreed with reverting #513 to improve it.

kenhys:
In this approach, the border case should also follow in CI.

kenhys:
Reverted #513.

lib/fluent/plugin/rdkafka_patch/0_14_0.rb
kenhys: Applied.

ashie (approved):
LGTM

ashie:
Thanks for your contribution! @ttych

PR #507, opened 2024-02-26
Update in_kafka_group.rb to fix the memory leak.

srikanthbirada:
In kafka group plugin is leaking memory incase of kafka is not reachable and disable retry limit is true.

### DIFF
--- a/lib/fluent/plugin/in_kafka_group.rb
+++ b/lib/fluent/plugin/in_kafka_group.rb
@@ -239,6 +239,7 @@ class Fluent::KafkaGroupInput < Fluent::Input
 
   def reconnect_consumer
     log.warn "Stopping Consumer"
+    # Following line is causing a memory leak, in scenario, where kafka is not reachable and disable_retry_limit is true.
     consumer = @consumer
     @consumer = nil
     if consumer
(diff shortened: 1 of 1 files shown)


lib/fluent/plugin/in_kafka_group.rb
```diff
@@ -254,6 +255,9 @@ def reconnect_consumer
     log.error "unexpected error during re-starting consumer object access", :error => e.to_s
     log.error_backtrace
     if @retry_count <= @retry_limit or disable_retry_limit
+      # Adding the following line to ensure, consumer object is cleaned before recursively calling reconnect_consumer.
+      # This ensures no memory leak.
+      consumer = nil
       reconnect_consumer
```
ashie, line 261: Probably just calling `retry` here is better than resetting `consumer` and calling `reconnect_consumer` recursively.

ashie (changes requested):
Sorry for my late response.

Could you please address following 2 points?
* Probably just calling`retry` is enough and better than calling `reconnect_consumer` recursively as I've commented in-line.
* Please follow DCO: https://github.com/fluent/fluent-plugin-kafka/pull/507/checks?check_run_id=21965716986
  (Please add `Signed-off-by: ` field to your commit message.)

PR #508, opened 2024-04-01
feat(out_waterdrop): add basic waterdrop output plugin

raytung:
This is a super simple proof-of-concept implementation for Waterdrop as described in https://github.com/fluent/fluent-plugin-kafka/issues/500. We'll need a lot more work to get this up to scratch compared to the other output plugins but I don't want to make this PR even larger :smile:


cosmo0920:
This could be included in v0.19.3 release. But can we mark as v0.20.0.rc milestone? This is because we have to seek successor of pure Ruby client for Kafka.

ashie:
IMHO there is no problem merging this in any timing with removing `gem.add_dependency 'waterdrop', '~> 2.6'` for make it as prototype of the new plugin.
Merging PoC code for new feature but disable by default is a common practice in large OSS projects such as Chromium or Firefox.

PR #512, opened 2024-07-29
Update target ruby versions

kenhys:
Drop deprecated ruby version

    * 3.0: EOL 2024-04-23
    * 2.7: EOL 2023-03-31

ref. https://www.ruby-lang.org/en/downloads/branches/

Then add Ruby 3.3 for test matrix

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


kenhys:
* ruby 3.1 rdkafka >= 0.12 crashes, but it was not regression in this PR.
* ruby 3.2 rdkafka >= 0.12 also crashes, but undefined method 'close' is same as before.

kenhys:
Fluentd still enables ruby 2.7 CI.

PR #513, opened 2024-07-29
Apply PR#468 patch only for 0.12.x.

kenhys:
With https://github.com/fluent/fluent-plugin-kafka/pull/468,
 rdkafka 0.12.0 specific patch was introduced, but rdkafka's internal change
was reverted later in 0.13.x.

So this monkey patch should be effective only between 0.12.0 and 0.12.1 (only 2 versions were shipped in 0.12.x).
For 0.13.x or later, it seems harmful.

PR #515, opened 2024-07-30
Add Ruby 3.3 for test matrix

kenhys:


kenhys:
As #513 was merged, then let's add 3.3 for test matrix now.

ashie (approved):

Issue #311, opened 2020-01-17
Unable to write after restarting kafka cluster


xidiandb:
When I restarted the kafka cluster, the following problems occurred:
```
2020-01-17 07:29:35 +0000 [warn]: #0 failed to flush the buffer. retry_time=10 next_retry_seconds=2020-01-17 07:30:33 +0000 chunk="59c39b45987d2889fdf53f5e2d87f58c" error_class=Kafka::UnknownTopicOrPartition error="Kafka::UnknownTopicOrPartition"
  2020-01-17 07:29:35 +0000 [warn]: #0 suppressed same stacktrace
2020-01-17 07:29:45 +0000 [warn]: #0 Send exception occurred: Kafka::UnknownTopicOrPartition
2020-01-17 07:29:45 +0000 [warn]: #0 Exception Backtrace : /var/lib/gems/2.5.0/gems/ruby-kafka-0.6.8/lib/kafka/protocol.rb:130:in `handle_error'
/var/lib/gems/2.5.0/gems/ruby-kafka-0.6.8/lib/kafka/protocol/metadata_response.rb:144:in `partitions_for'
/var/lib/gems/2.5.0/gems/ruby-kafka-0.6.8/lib/kafka/cluster.rb:155:in `partitions_for'
/var/lib/gems/2.5.0/gems/fluent-plugin-kafka-0.7.9/lib/fluent/plugin/kafka_producer_ext.rb:190:in `assign_partitions!'
/var/lib/gems/2.5.0/gems/fluent-plugin-kafka-0.7.9/lib/fluent/plugin/kafka_producer_ext.rb:153:in `block in deliver_messages_with_retries'
/var/lib/gems/2.5.0/gems/fluent-plugin-kafka-0.7.9/lib/fluent/plugin/kafka_producer_ext.rb:148:in `loop'
/var/lib/gems/2.5.0/gems/fluent-plugin-kafka-0.7.9/lib/fluent/plugin/kafka_producer_ext.rb:148:in `deliver_messages_with_retries'
/var/lib/gems/2.5.0/gems/fluent-plugin-kafka-0.7.9/lib/fluent/plugin/kafka_producer_ext.rb:102:in `deliver_messages'
/var/lib/gems/2.5.0/gems/fluent-plugin-kafka-0.7.9/lib/fluent/plugin/out_kafka2.rb:220:in `write'
[...]

PR #519, opened 2024-09-29
kafka2 bugfix: Failed to assign partitions

leo198706:
1. fluent-plugin-kafka 0.18.1
2. fluentd 1.14.6

[error.log](https://github.com/user-attachments/files/17177426/error.log)

According to the above error, I analyzed the reasons as follows：

When Kafka does not have a topic to write to and does not allow automatic creation, an error UnknownTopicOrPartition will be reported, but it will be caught by Kafka::Error 
https://github.com/fluent/fluent-plugin-kafka/blob/v0.18.1/lib/fluent/plugin/kafka_producer_ext.rb#L254

and finally thrown in rescue Exception => e. 
https://github.com/fluent/fluent-plugin-kafka/blob/master/lib/fluent/plugin/out_kafka2.rb#L411

This is not an ideal situation.

So I adjusted the code to handle the above situation and avoid repeated retries.


JensErat:
Is this related to https://github.com/fluent/fluent-plugin-kafka/issues/311 ?

Watson1978:
We think the behavior you reported is not bug.

To avoid repeated retries, you might use `discard_kafka_delivery_failed` option.

PR #526, opened 2025-02-06
github: add Ruby 3.3/3.4 for CI

kenhys:

PR #528, opened 2025-03-04
ci: use bundler installed by ruby/setup-ruby action

Watson1978:
ruby/setup-ruby action installs the appropriate bundler for each Ruby version.
If the combination is inappropriate, it may not work as expected.

Related to https://github.com/fluent/fluent-plugin-kafka/pull/526#issuecomment-2639114023


kenhys (approved):
LGTM.

Issue #536, opened 2025-09-11
rdkafka_group don't support  regex pattern topics


wcollin:
### Describe the bug

rdkafka_group don't support  regex pattern topics

### To Reproduce

i use fluent/fluentd-kubernetes-daemonset:v1.19-debian-kafka2-1

### Expected behavior

topics supports regex pattern like "@type kafka_group"

### Your Environment

```markdown
- Fluentd version: 1.19.0
- fluent-plugin-kafka version: 0.19.5
- ruby-kafka version: 0.22.2 x86_64-linux-gnu
- rdkafka version: 1.5.0
```

### Your Configuration

```
<source>
  @type rdkafka_group
  @log_level info
  @label @toes

  kafka_configs {
    "bootstrap.servers": "#{ENV['FLUENTD_KAFKA_ADDR']}",
    "group.id": "logcenter_group",
    "auto.commit.interval.ms": "10000",
    "topic.metadata.refresh.interval.ms": "60000"
  }

  topics /test-*/
  add_prefix "kafka"
  format json
  retry_emit_limit 10

  # kafka consumer options
  max_wait_time_ms 500
  max_batch_size 10000
</source>
```

### Your Error Log

```shell
2025-09-11 08:56:56 +0000 [error]: #0 unexpected error during consuming events from kafka. Re-fetch events. error="Broker: Unknown topic or partition (unknown_topic_or_part)"
  2025-09-11 08:56:56 +0000 [error]: #0 /fluentd/vendor/bundle/ruby/3.4.0/gems/rdkafka-0.22.2-x86_64-linux-gnu/lib/rdkafka/consumer.rb:550:in 'Rdkafka::Consumer#poll'
  2025-09-11 08:56:56 +0000 [error]: #0 /fluentd/vendor/bundle/ruby/3.4.0/gems/fluent-plugin-kafka-0.19.5/lib/fluent/plugin/in_rdkafka_group.rb:209:in 'Fluent::Plugin::RdKafkaGroupInput#each_batch'
  2025-09-11 08:56:56 +0000 [error]: #0 /fluentd/vendor/bundle/ruby/3.4.0/gems/fluent-plugin-kafka-0.19.5/lib/fluent/plugin/in_rdkafka_group.rb:229:in 'Fluent::Plugin::RdKafkaGroupInput#run'
[...]

PR #541, opened 2025-10-22
in_rdkafka_group: support regexp pattern in topics

Watson1978:
https://github.com/fluent/fluent-plugin-kafka/issues/536

This PR will add regexp pattern syntax to subscribe the multiple topics like in_kafka_group.

### DIFF
--- a/lib/fluent/plugin/in_rdkafka_group.rb
+++ b/lib/fluent/plugin/in_rdkafka_group.rb
@@ -70,18 +70,28 @@ class Fluent::Plugin::RdKafkaGroupInput < Fluent::Plugin::Input
   end
 
   def _config_to_array(config)
-    config_array = config.split(',').map {|k| k.strip }
+    config_array = config.split(',').map {|k| _config_regex_pattern(k.strip) }
     if config_array.empty?
       raise Fluent::ConfigError, "kafka_group: '#{config}' is a required parameter"
     end
     config_array
   end
+  private :_config_to_array
+
+  def _config_regex_pattern(topic)
+    if (m = /^\/(.+)\/$/.match(topic))
+      # librdkafka recognizes string as regex pattern if the topic name starts with '^'.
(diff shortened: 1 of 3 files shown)


test/plugin/test_in_rdkafka_group.rb
```diff
@@ -0,0 +1,123 @@
...
+        </parse>
+      ]
+      d = create_driver(conf)
+
+      d.run(expect_records: 2, timeout: 10) do
+        sleep 0.1
+        @producer.produce("Hello, fluent-plugin-kafka! in topic 1", topic: TOPIC_NAME1)
+        @producer.produce("Hello, fluent-plugin-kafka! in topic 2", topic: TOPIC_NAME2)
+        @producer.deliver_messages
+      end
```
kenhys, line 121: Wouldn't it be better to include test data on topics that didn't match.
Then test for those cases as well?

kenhys (changes requested):

lib/fluent/plugin/in_rdkafka_group.rb
```diff
@@ -70,19 +70,29 @@ def initialize
   end
 
   def _config_to_array(config)
-    config_array = config.split(',').map {|k| k.strip }
+    config_array = config.split(',').map {|k| _config_regex_pattern(k.strip) }
```
Watson1978: Current implementation uses a simple config.split(',') to separate topics.
This will incorrectly break regex patterns that contain commas within the pattern itself like `topics /custom-topic-[0-9]{1,3}/`

Need to fix this.

Watson1978: Fixed with https://github.com/fluent/fluent-plugin-kafka/pull/541/changes/f275bb41c1cfc0e8813c7a52f7e39abd7908e480

test/plugin/test_in_rdkafka_group.rb
Watson1978: Thanks.

Added in https://github.com/fluent/fluent-plugin-kafka/pull/541/changes/d3159bf0afc3d79232658b62766c5f86cc89e89b

kenhys (approved):

PR #562, opened 2026-06-02
ci: update baseline of confluent

kenhys:
https://docs.confluent.io/platform/current/installation/versions-interoperability.html#cp-and-apache-ak-compatibility It might be better to update supported one.

EOL of 7.8.x: December 2, 2026.


Watson1978 (approved):
👍🏻

PR #566, opened 2026-06-10
Replace yajl-ruby with the json gem for JSON handling

Watson1978:
Since the yajl-ruby gem depends on a deprecated Ruby C API, it cannot be installed with Ruby 4.1.


kenhys (approved):
LGTM

